PR #17, opened 2017-08-04
Settings page

giacomolicari:
This feature manages key=>value settings item declaration.

We found a limitation:
- We would like to save elements to redux-store only when the form doesn't have any error. There's a validation function inside the settings' component file that detects those errors, however, we couldn't get it working inside the onChange() function (it may be an asynchronous problem)


src/containers/SettingsPage/index.js
```diff
@@ -1,5 +1,43 @@
...
 
-export default connect()(Settings)
+const FORM = {
+  form: 'settingsForm',
+  onChange: async (values, dispatch, props) => {
+    const { formValues, syncErrors, pristine, dirty } = props
+    // TODO should only if no errors but seems to be an issue with
+    // onchange validation
+    // https://github.com/erikras/redux-form/issues/3012
+    return await dispatch(updateSettings(formValues.settings))
```
andre-meyer: Instead of validating onChange you can instead validate on "onSubmit" and change this handler to immediately submit the form, then you can use the validator function here.

(I think `.submit()` exists in props, otherwise import the action creator to submit a form.)

src/reducers/settings.js
```diff
@@ -0,0 +1,19 @@
...
+  [updateSettings]: (state, action) => {
+    console.log('REDUX ', state)
+    const data = action.payload
+    let result = {}
+    data.map((item) => {
+      if (item.address && item.name) {
+        result[item.address] = item.name
+      }
+    })
+    return result
```
andre-meyer: I would like to already nest the mappings here within the settings. I think we're going to use settings for more than just the whitelist mapping later on, maybe it would be better if the reducer wrote into `settings.mapping`? On Line 17 you can initialize the reducer with an empty object as the mapping (`, { mapping: {} }`)

src/components/Settings/index.js
```diff
@@ -1,10 +1,113 @@
...
+    // Value Errors
+    if ( occurrences.length > 1 ) {
+      error.address = 'Duplicated Address'
+      settingsErrors[index] = error
+    }
+    else if (!item.address) {
+      error.address = 'Required'
+      settingsErrors[index] = error
+    }
+    else if (item.address.length != 42 || item.address.substr(0, 2) != '0x') {
```
andre-meyer: Instead of requiring the 0x prefix, we can also automatically prefix it with the normalize and parse props on the field: 
- http://redux-form.com/7.0.3/docs/api/Field.md/#-code-parse-value-name-gt-parsedvalue-code-optional-
- http://redux-form.com/7.0.3/docs/api/Field.md/#-code-normalize-value-previousvalue-allvalues-previousallvalues-gt-nextvalue-code-optional-

If this doesn't work as easily feel free to skip it for now but it would be nice to have

src/components/Settings/index.js
```diff
@@ -1,10 +1,113 @@
...
+  const settingsErrors = []
+  for (var index=0; index<values.settings.length; index++) {
+    const error = {}
+    const item = values.settings[index]
+    // Name errors
+    if (!item.name || item.name.length < 3) {
+      error.name = 'Required'
+      settingsErrors[index] = error
+    }
+    const occurrences = values.settings.filter((other) => {return item.address == other.address})
```
andre-meyer: This can be written as 
`(other) => item.address == other.address`

Should be a linter error I think

src/components/Settings/index.js
```diff
@@ -1,10 +1,113 @@
...
+        </div>
+      )}
+      </div>
+    </div>
+
+// TODO move validate functions to a generic utilities file
+const validate = values => {
+  const errors = {}
+  const settingsErrors = []
+  for (var index=0; index<values.settings.length; index++) {
```
andre-meyer: Prefer `.forEach` here or change `var` to `let`.

src/containers/SettingsPage/index.js
andre-meyer: Oh wait, "updateSettings" already submits the form. Check the [onSubmit](http://redux-form.com/7.0.3/docs/api/ReduxForm.md/#-code-onsubmit-function-code-optional-) function if you can access the validator errors there!

PR #24, opened 2017-08-14
Feature/my trades

giacomolicari:


src/components/MarketDetail/index.js
```diff
@@ -195,8 +206,8 @@ class MarketDetail extends Component {
       <div className="marketControls container">
         <div className="row">
           {Object.keys(expandableViews).filter(view =>
-            !expandableViews[view].showCondition ||
-            expandableViews[view].showCondition(this, market),
+            typeof expandableViews[view].showCondition !== 'function' ||
+            expandableViews[view].showCondition(this.props, market),
```
andre-meyer: second parameter here is not used anymore since we pass the whole props

src/components/MarketMyTrades/index.js
```diff
@@ -0,0 +1,102 @@
+import React, { Component } from 'react'
+import Decimal from 'decimal.js'
+
+// import { calcLMSRMarginalPrice, calcLMSROutcomeTokenCount } from 'api'
```
andre-meyer: remove this please

src/components/MarketMyTrades/index.js
```diff
@@ -0,0 +1,102 @@
...
+  getAverageCost(order) {
+    if (order.orderType == 'BUY') {
+      return order.cost/1e18
+    }
+    else if (order.orderType == 'SELL') {
+      return order.profit/1e18
+    }
+    else if (order.orderType == 'SHORT SELL') {
+      return order.cost/1e18
+    }
```
andre-meyer: this should be a linter error, I think you need to explicitly return a value if none of these cases are true

src/components/MarketMyTrades/index.js
```diff
@@ -0,0 +1,102 @@
...
+            </tbody>
+          </table>
+        </div>
+      )
+    }
+    else {
+      return (
+        <div className="marketMyTrades">
+          <h2 className="marketMyTrades__heading">You haven't interacted with this market yet.</h2>
+          <h3 className="marketMyTrades">Every transaction that happens on this market will be shown here.</h3>
```
andre-meyer: Why is this the same class as the block?

src/components/MarketMyTrades/marketMyTrades.less
```diff
@@ -0,0 +1,88 @@
...
+  &__shareTable {
+    margin-top: 20px;
+    table-layout: fixed;
+  }
+
+  &__share {
+    vertical-align: middle;
+
+    td {
+      //border-top: 0 !important;
```
andre-meyer: remove this please

src/utils/helpers.js
```diff
@@ -42,6 +42,17 @@ export const toEntity = (data, entityType, idKey = 'address') => {
...
 
+/** 
+ * Adds _id incremental numeric property to each object in the array
+ * @param {Array of objects} arrayData 
+ */
+export const addIdToObjectsInArray = (arrayData) => {
+  arrayData.forEach((item, index) => {
+    item['_id'] = index
+  })
+  return arrayData
```
andre-meyer: Very cool to move this logic here, we reuse this a lot 👍

andre-meyer:
I think you accidentally kept your changes from the Shortsell branch too. Next time be sure to checkout from development before you branch for a new feature

src/components/MarketDetail/index.js
andre-meyer: If this is also used at the other place where we check for the showCondition, remove it there too please

giacomolicari:
Code updated

PR #30, opened 2017-08-21
Feature/polish settings page

vianneylancres:


src/components/Settings/settings.less
```diff
@@ -0,0 +1,239 @@
...
+//   letter-spacing: .5px;
+// }
+
+// .marketControls.container {
+//   padding-top: 40px;
+// }
+
+// .marketControls__button {
+//   margin-right: 20px;
+// }
```
andre-meyer: Can you remove these comments here? If they're not needed anymore, just remove them. Commenting out code can be avoided because we use git :)

PR #35, opened 2017-08-28
Task/polish market overview page

denisgranha:
- Add radiobutton Resolved Unresolved All
- Remove oracle type filter
- Add sort by Section above marketlist (dropdown): Resolution Date, Trading volume, asc/desc is fix
- Don't show oracle information
- Add trading volume (part of gnosisdb response)
- Add Checkbox to filter to show only my markets


src/components/FormSelect/index.js
```diff
@@ -6,17 +6,18 @@ import { pick } from 'lodash'
 import { bemifyClassName } from 'utils/helpers'
 
 const FormSelect = ({ input, label, values, className, defaultValue, ...props }) => {
-  const selectProps = pick(props, [])
-
+  const selectProps = pick(props, [])
```
andre-meyer: What does this do?

denisgranha: nothing really, just a removed space, I guess because there was a debugger placed on those lines, and after debugging we get rid of it ^^
selectedProps was there already on the develop branch

PR #39, opened 2017-09-05
fixed bug, doesn't allowed to buy shares after creating a market with…

denisgranha:
…out reloading the page


andre-meyer:
Good find and good fix!

PR #44, opened 2017-09-07
Bug/fix share earnings

denisgranha:
Depends on branch task/use-gnosisjs-estimations


src/components/MarketMySharesForm/index.js
```diff
@@ -27,6 +27,17 @@ class MarketMySharesForm extends Component {
     }
   }
 
+  componentWillMount() {
+    const { gasCosts, gasPrice, requestGasCost, requestGasPrice } = this.props
+
+    if (gasCosts.sellShares === 0) {
+      requestGasCost(GAS_COST.SELL_SHARES)
+    }
+    if (gasPrice.eq(0)) {
```
andre-meyer: The default should be undefined in the reducer and request the data if it's not available. Inside the components it should fallback to 0 if the gas cost is undefined.

Because if we actually have gas costs that are 0, this would turn into an infinite loop.

src/utils/helpers.js
```diff
@@ -48,8 +48,12 @@ export const toEntity = (data, entityType, idKey = 'address') => {
  */
 export const weiToEth = (value) => {
   let ethValue = '0'
-  if (value && Decimal(value).gt(0)) {
-    ethValue = Decimal(value).div(1e18).toString()
+  if (value) {
+    if (value.gt && value.gt(0)) {
```
andre-meyer: Hmm, if the value has the `.gt`method but it is not greater than 0, it will try to parse it to a new decimal? What if it is a negative number?

To typecheck here for decimal I'd rather use `typeof value === 'string'` and then parse to decimal

src/selectors/blockchain.js
```diff
@@ -1,9 +1,13 @@
-import { get } from 'lodash'
+import Decimal from 'decimal.js'
 
 export const selector = state => state.blockchain
 
 export const getDefaultAccount = state => selector(state).defaultAccount
 
+export const getGasCosts = state => selector(state).gasCosts
+
+export const getGasPrice = state => new Decimal(parseInt(selector(state).gasPrice, 10))
```
andre-meyer: Same here as above, I'd prefer it if we returned undefined here, for gas prices that have not been set yet

andre-meyer:
Oops. I think I did the wrong branch again 😄 Next time you can create the PR with the dependant branch as the base. That way I only see the changes you did in your branch in relation to the depedant branch

src/components/MarketMySharesForm/index.js
denisgranha: is set default to 0 to avoid checks in the cost calculations, but yeah, it opens the door to infinite loop in case of gnosis.js returns 0 for any of the costs
This case should never happen, gnosis.js takes costs from a generated json file

src/utils/helpers.js
denisgranha: right

src/components/MarketMySharesForm/index.js
andre-meyer: Maybe we can have a seperate method in the selector that determines if the gas cost was already fetched (e.g. not `undefined`) and a second selector that returns the gas cost and if it does not exist, returns 0 or another default

andre-meyer: That way you can avoid having to check/cast everywhere before using the gas cost values while still being able to tell with more accuracy, if the values were already fetched.

This could also be solved on a reducer level by introducing new variables that indicate if the gas cost were fetched. What do you think?

PR #46, opened 2017-09-12
Audit/market overview

giacomolicari:


src/containers/MarketListPage/index.js
```diff
@@ -9,6 +9,8 @@ import { getDefaultAccount } from 'selectors/blockchain'
 
 import { requestMarkets } from 'actions/market'
 
+const config = require('config.json')
```
andre-meyer: why not import?

src/components/MarketList/index.js
```diff
@@ -121,14 +121,14 @@ class MarketList extends Component {
...
-          <div className="info__group col-md-3">
+          {/* <div className="info__group col-md-3">
             <div className="info__field">
               <div className="info__field--icon icon icon--currency" />
               <div className="info__field--label">
                 <CurrencyName collateralToken={market.event.collateralToken} />
               </div>
             </div>
-          </div>
+          </div>*/}
```
andre-meyer: Can you remove commented out sections please? They will be in our git log if they're ever needed again

src/containers/MarketListPage/index.js
giacomolicari: config.json contains a JSON dictionary, not a Javascript object. Is there a way to use Import with json files?

andre-meyer: What happens if you try `import config from 'config.json'`?

PR #52, opened 2017-09-18
Task/package updates and deprecated babel presets removal

mmvsha:
Update few packages, updated .babelrc config. For me build seems to be stable. 
screenshot from 2017-09-18 15 46 33 this is what else we need to update


denisgranha:
Have you tested to:
* Create market
* Buy Shares
* Sell shares
* See Market list
* Market graph is rendered correctly 

@mmvsha

denisgranha:
ok, we tested the features and it worked

PR #54, opened 2017-09-19
webpack config optimizations

mmvsha:
I've split the webpack configurations, a bit optimized both of them, now its a bit faster build speed. We need to discuss what to do with the routes because if we keep it done like this we cannot use react-hot-loader


package.json
```diff
@@ -59,6 +59,7 @@
     "redux-thunk": "^2.2.0",
     "reselect": "^2.5.4",
     "sha1": "^1.1.1",
+    "truffle-contract": "^3.0.0",
```
andre-meyer: Can you remove this change? I'm afraid it will break something if we simply deploy with it now

andre-meyer:
Very Nice! 👍 But let's please wait until we can adjust the deployment for this.

Regarding your changes that need to be done before we can have react hot reloading:

I've added a ticket in trello in the backlog for it, when you find the time feel free to split it up. I'd suggest we also change the main "App.js" part and the AppProvider Section into two files as well, so that we can completely seperate the hot reloading from the remaining application.

https://trello.com/c/wGPuYsnH/154-adjust-routes-entry-for-react-hot-reloading

package.json
mmvsha: @andre-meyer yes, sure. My local dev build wouldn't compile without this for some reason, thought i had to add it.

src/router.js
```diff
@@ -0,0 +1,55 @@
...
+class AppRouter extends Component {
+  static propTypes = {
+    history: PropTypes.object,
+  }
+
+  render() {
+    const { history } = this.props
+
+    return (
+      <Router key={Math.random()} history={history}>
```
andre-meyer: Why the `math.random()` key?

webpack.dev.config.js
```diff
@@ -0,0 +1,116 @@
...
+const webpack = require('webpack')
+const pkg = require('./package.json')
+
+const nodeEnv = process.env.NODE_ENV || 'development'
+const version = process.env.BUILD_VERSION || pkg.version
+const build = process.env.BUILD_NUMBER || 'SNAPSHOT'
+
+const config = require('./src/config.json')
+
+// const ethereumHost = process.env.ETHEREUM_HOST
```
andre-meyer: Not needed anymore? If not, remove it please

webpack.dev.config.js
```diff
@@ -0,0 +1,116 @@
...
+    contentBase: false,
+    port: 5000,
+    proxy: {
+      '/api': {
+        target: gnosisDbUrl,
+        secure: false,
+      },
+    },
+    watchOptions: {
+      ignored: /node_modules/,
```
andre-meyer: Often enough I have to link my local gnosis.js package and when I compile contracts from there, it should rebuild. Can we remove this, or is this a huge deal for performance/bundlesize?

webpack.dev.config.js
```diff
@@ -0,0 +1,116 @@
...
+        windows: false,
+      },
+    }),
+    new HtmlWebpackPlugin({
+      template: path.join(__dirname, 'src/html/index.html'),
+    }),
+    new webpack.DefinePlugin({
+      'process.env': {
+        VERSION: JSON.stringify(`${version}#${build}`),
+        // ETHEREUM_HOST: nodeEnv === 'production' ? null : JSON.stringify(ethereumHost),
```
andre-meyer: Same here, remove if it it's not necessary

webpack.prod.config.js
```diff
@@ -13,15 +13,12 @@ const build = process.env.BUILD_NUMBER || 'SNAPSHOT'
...
 
 // const ethereumHost = process.env.ETHEREUM_HOST
-const gnosisDbUrl = process.env.GNOSISDB_HOST || `${config.gnosisdb.protocol}://${config.gnosisdb.host}:${config.gnosisdb.port}`
+const gnosisDbUrl =
+  process.env.GNOSISDB_HOST || `${config.gnosisdb.protocol}://${config.gnosisdb.host}:${config.gnosisdb.port}`
 
 module.exports = {
   context: path.join(__dirname, 'src'),
-  entry: [
-    'bootstrap-loader/extractStyles',
```
andre-meyer: Why not extractStyles for prod?

mmvsha: @andre-meyer extractStyles can be specified by two ways:
1. with entry like this
2. Add a parameter in .bootstraprc
We had it specified both ways and to avoid duplication I decided to remove it from config and leave it in .bootstraprc

webpack.dev.config.js
mmvsha: @andre-meyer Its done for webpack-dev-server so it doesn't watch for changes in node_modules folder. I had problems with react-hot-loader not working because node_modules is large and to avoid checking all node_modules large tree it decides to recompile the app instead of HMR

src/router.js
mmvsha: @andre-meyer https://github.com/reactjs/react-router-redux/issues/179

webpack.dev.config.js
andre-meyer: Alright! 👍

PR #58, opened 2017-09-22
Fixed bug when deleting letters on outcome

giacomolicari:


src/components/SectionOutcomeCategorical/index.js
```diff
@@ -11,9 +11,6 @@ const OutcomeCategorical = () => (
           name="outcomes"
           component={OutcomeList}
           label="Outcomes"
-          validate={(val) => {
-            return val && val.filter(v => v.length > 0).length > 0 ? undefined : 'Field is required'
-          }}
```
andre-meyer: How did you test this? I think the `required` validator will allow empty entries like this: `['Outcome A', 'Outcome B', '']`

andre-meyer: It should also have atleast 2 entries, right? Maybe we should add that validation as well.d

If the required validator does allow empty entries, we could also just normalize it before we save, by filtering out all empty entries: https://redux-form.com/7.0.4/docs/api/field.md/#-code-normalize-value-previousvalue-allvalues-previousallvalues-gt-nextvalue-code-optional-

giacomolicari: Yes, I re-check this.

giacomolicari: Required validator requires the input string to have more than 0 chars, so we would add min-length validator as well.

PR #62, opened 2017-09-25
Declared api url in config

giacomolicari:


mmvsha:
@giacomolicari Why declare it in api config if we can just set up environment variable? like process.env.API_URL

giacomolicari:
@mmvsha I'm going to remove apiUrl and use GnosisDB configuration properties from config.json.
Another approach would be using environment variables as you said. We can start using config.json first and introduce environment configuration variables afterwards.

PR #66, opened 2017-09-28
add modal warning for no account, change loader

mmvsha:


src/components/MarketProgress/index.js
```diff
@@ -14,11 +14,12 @@ const MarketProgress = ({
...
-  
-  if (failed) {
-    strokeDasharray = 0
-  }
+  // its not used
+  // let strokeDasharray = Math.abs((progress) - 1) * Math.PI * (190 * 2)
+
+  // if (failed) {
+  //   strokeDasharray = 0
+  // }
```
andre-meyer: Feel free to remove it, this is from `ProgressSpinner`, I forgot to remove it

src/containers/App/index.js
```diff
@@ -31,20 +37,37 @@ class App extends Component {
...
+          </CSSTransition>
+        </TransitionGroup>
+        <Modal
+          isOpen={this.props.blockchainConnection && !this.props.account}
+          contentLabel="no-account-modal"
+          style={modalStyles}
+        >
+          <h1 id="heading">Oops!</h1>
+          <div id="description">
+            <p>We couldn't detect your account. Please check your wallet provider and reload the page</p>
```
andre-meyer: Can we fix this problem in Github where the syntax always messes up? That's btw why I always escape `'`

src/components/Icons/loading.js
```diff
@@ -0,0 +1,36 @@
...
+        from="0 20 20"
+        to="360 20 20"
+        dur="0.5s"
+        repeatCount="indefinite"
+      />
+    </path>
+  </svg>
+)
+
+export default LoadingIcon
```
andre-meyer: Can we use the `ProgressSpinner` instead maybe? Or is this completely different? I want to keep the loading indicators similar

andre-meyer: Ah, the Loading Icon is more for undetermined length. Maybe you can add that to the ProgressSpinner, or create a wrapper that handles it.

mmvsha: @andre-meyer Should I? `ProgressSpinner` is a component for showing progress, and this is just a wrapper on svg icon

andre-meyer: But it's a loading indicator kind of thing right? Maybe you can reuse some of it so it looks similar?

PR #74, opened 2017-10-03
Bug/show close market button only for creator

mmvsha:


src/components/MarketDetail/index.js
```diff
@@ -257,38 +275,46 @@ class MarketDetail extends Component {
...
+                  ? 'marketControls__button--active btn btn-primary'
+                  : expandableViews[view].className}`}
+                onClick={() => this.handleExpand(view)}
+              >
+                {expandableViews[view].label}
+              </button>
+            ))}
+          {market.stage !== MARKET_STAGES.MARKET_CLOSED &&
+          market.creator === defaultAccount &&
+          process.env.WHITELIST[defaultAccount] ? (
```
andre-meyer: `process.env.WHITELIST` is this not inside the config.json? Can that be used instead?

giacomolicari: We needed to have two kind of whitelists, one for production and one for development with the list of the default TestRPC accounts.
In the webpack config we check for the environment type (dev or prod) and pick productionWhitelist or developmentWhitelist:
https://github.com/gnosis/gnosis-management/blob/master/webpack.dev.config.js#L16

andre-meyer: Ah, nevermind. It's because of the dev/prod switch. Got it!

PR #75, opened 2017-10-03
fix already resolved markets showing up in soon-closing markets widget

mmvsha:


src/components/Dashboard/index.js
```diff
@@ -287,7 +287,9 @@ class Dashboard extends Component {
     /* const closingMarkets = markets.filter(
       market => moment.utc(market.eventDescription.resolutionDate).isBetween(moment(), moment().add(24, 'hours')),
     )*/
-    let closingMarkets = markets.sort((a, b) => a.eventDescription.resolutionDate > b.eventDescription.resolutionDate)
+    let closingMarkets = markets
+      .filter(market => new Date() - new Date(market.eventDescription.resolutionDate) < 0)
```
andre-meyer: Hmm, the `resolutionDate` is not a really a strict time when the market will be closed. The important thing here is if the outcome was set or not. For example if I set my Market to resolve tomorrow at 10:00 AM and I somehow accidentally delay it by 20 minutes, it should still show up here until the outcome was set.

andre-meyer: I know this is already closed, but what do you think @denisgranha @giacomolicari ?

PR #77, opened 2017-10-04
Documentation: Add JSDocs to MarketActions and Models

andre-meyer:


giacomolicari (approved):

mmv08 (approved):
Looks good, but is it necessarily to write "This function is asynchronous." before function with "async" keyword?

andre-meyer:
Probably not @mmvsha, I removed it

PR #80, opened 2017-10-05
Add source maps in webpack configuration

apanizo:
DESCRIPTION
This PRs introduces source-maps for being used with Chrome DevTools only when developing.

TESTS DONE
I have tested the app running in bash
```
NODE_ENV=development npm start
NODE_ENV=production npm start
```

Checked both versions run the correct scripts and loaded the app.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -15,5 +15,6 @@
         "transform-react-remove-prop-types"
       ]
     }
-  }
+  },
+  "sourceMaps": true
 }
--- a/.babelrc
+++ b/.babelrc
@@ -15,6 +15,5 @@
         "transform-react-remove-prop-types"
       ]
     }
-  },
-  "sourceMaps": true
+  }
 }
--- a/webpack.dev.config.js
+++ b/webpack.dev.config.js
@@ -11,14 +11,7 @@ const build = process.env.BUILD_NUMBER || 'SNAPSHOT'
 
 const config = require('./src/config.json')
 
-let whitelist
-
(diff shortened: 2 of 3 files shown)


webpack.dev.config.js
```diff
@@ -29,7 +22,7 @@ const ethereumUrl =
 module.exports = {
   context: path.join(__dirname, 'src'),
   entry: ['react-hot-loader/patch', 'bootstrap-loader', 'index.js'],
-  devtool: 'eval',
+  devtool: 'source-map',
```
mmvsha: https://webpack.js.org/configuration/devtool/
looks like source-map is not good for rebuilding, maybe switch to other one?

.babelrc
```diff
@@ -15,5 +15,6 @@
         "transform-react-remove-prop-types"
       ]
     }
-  }
+  },
+  "sourceMaps": true
```
Velenir: This setting is ignored when used with webpack. See **Note** at https://github.com/babel/babel-loader#user-content-options

webpack.dev.config.js
mmvsha: `eval-source-map`, for example

Velenir: eval-source-map may be a better fit. It is faster on rebuild and used for development only.

apanizo:
Updated!

.babelrc
apanizo: Good catch.

PR #81, opened 2017-10-05
fix bug when shares are not shown after buy

mmvsha:
**DESCRIPTION:**
- This PR fixes shares not showing on `My shares` tab after buying.

**TESTS:**
_Previously_: After buying shares they're not shown in `My shares` tab, had to update the page.
_Now_: After buying shares they're shown in `My shares` tab.
_Test_: Manually tested. Got desired behaviour.


apanizo:
Please, introduce a short description, screenshots of before and after, tests what you have done (manual ones).

src/components/MarketMySharesForm/index.js
```diff
@@ -93,18 +104,93 @@ class MarketMySharesForm extends Component {
...
     if (decimalValue.lt(0)) {
-      return 'Number can\'t be negative.'
+      return "Number can't be negative."
     }
 
-    if (decimalValue.gt(Decimal(props.marketShares.filter(
-      share => share.id === this.state.extendedSellId)[0].balance,
-    ).div(1e18).toString())) {
-      return 'You\'re trying to sell more than you invested.'
+    if (
```
apanizo: Refactor to something like:
```
const myUsefulDecimalName = Decimal(props.marketShares.filter(share => share.id ....).div(1e18).toString(),
if (decimalValue.gt(myUsefulDecimalName)) { 
```

src/components/MarketMySharesForm/index.js
```diff
@@ -256,82 +345,26 @@ class MarketMySharesForm extends Component {
...
-            </td>
-          </tr>
-        ))
-      }
-    })
-
-    return tableRows
-  }
-
   render() {
```
apanizo: Refactor this render method creating private components for each situation. Having more than one return in the render method is confusing.
```
const hasShares = !marketShares || !marketShares.length;
return hasShares ? <SharesMarket /> : <EmptyMarket />
```
(can't find better names :) )

When I say Private components I mean something like: 
```
const EmptyMarket = () => <div ...>
    ....
</div>

//SharesMaker I would create this component as a functional component in ./SharesMarket otherwise this component is huge 400 lines is too much.

class MarketMySharesForm extends Component { 
```

src/components/MarketMySharesForm/index.js
```diff
@@ -359,9 +392,7 @@ class MarketMySharesForm extends Component {
               <th className="marketMyShares__tableHeading marketMyShares__tableHeading--group" />
             </tr>
           </thead>
-          <tbody>
-            {this.generateTableRows()}
-          </tbody>
+          <tbody>{this.generateTableRows()}</tbody>
```
apanizo: This should be also a component, functions should be avoided (for each render we generate a new reference in the stack).

If you need to calculate some values, those calculations should be done in a container of the main component (MarketMySharesForm).

This sub-component should look like:
```
const SharesRows = (firstCalculation, secondCalculation, ...) => <div>
    ... // other necessary html
    { numberOfShares.map( (share) => <div>
        ....
     </div> }
```

src/components/MarketMySharesForm/index.js
mmvsha: @apanizo I completely agree that it should be refactored, but i haven't changed those components. I've only fixed eslint errors and formatting in those files.

src/components/MarketMySharesForm/index.js
andre-meyer: @apanizo Can you elaborate please? What do you mean with we generate a new reference in the stack?
Afaik the performance hit between a function call and a stateless component is about equal, if not worse. As it would improve our markup, I can still agree with this, I'm just curious about what performance benefits this would gain us, as I've used function calls to render smaller parts of a react component a lot

PR #84, opened 2017-10-10
Feature: Live Refreshing Wallet

andre-meyer:
# Description:

### `WalletIntegrationProvider` Refactor
I've improved the functionality of the wallet provider layer and implemented a 'watcher' system to all integrations. Currently it's not supported to listen to any kind of event on a web3 Provider so we're instead updating every second, if something about the integration has changed. (Currently: Network, Account or Balance)

### Provider Priority System
Previously, if all other wallet providers failed to connect, the `REMOTE` provider would be used instead. I've now added a var called `providerPriority` to all integration, which enables us to selectively prefer integrations over others. (For example prefer to use Parity even if Metamask is available because Parity is a dedicated browser)

### Current Provider Information in Redux Store
[See here for quick demo](https://i.imgur.com/54QFx1Z.gifv)

The currentProvider is now a key in the blockchain redux state. The selectors have been changed so only the currentProvider Information is displayed. Inside the Header we now display the Network if it's not the main network. Provider Icon Alt-Text is the Providers Name if the Icon is not clear. Identicon Alt-Text is Account Address.

#### Changed functionality:
- Modal is now only visible if no connection could be made to any sort of wallet. So either connectivity problems or blocked requests. (I know we'll alter this anyway according to new specs)

# Testing:
- Change your Metamask User and watch the Identicon Update
- Change your Network or Account and watch the Balance change
- Change your Provider or disable your current one and see the Icon update
- Lock your Metamask and see that the current Information will disappear
- Inside the Redux State the currentProvider should be set after all providers have (tried) to initialized


src/actions/blockchain.js
```diff
@@ -30,39 +30,6 @@ export const setEtherTokens = createAction('SET_ETHER_TOKENS')
 
 const NETWORK_TIMEOUT = process.env.NODE_ENV === 'production' ? 10000 : 2000
```
mmvsha: Do you think 10 seconds is enough? I think in production we want it to be very fast

src/components/Header/header.less
```diff
@@ -107,18 +108,34 @@
   }
 }
 
-.headerContainer__account {
-  display: inline-block;
+.headerContainer {
+
+  &__network {
+    display: inline-block;
```
mmvsha: I was not able to find an element that will match `.headerContainer__network` selector, so probably we do not need `display` property here

src/components/Header/header.less
```diff
@@ -107,18 +108,34 @@
   }
 }
 
-.headerContainer__account {
-  display: inline-block;
+.headerContainer {
```
mmvsha: also we already have `.headerContainer` styles at line 5

src/components/Header/header.less
```diff
@@ -107,18 +108,34 @@
...
-    font-weight: 500;
-    padding-right: 5px;
-    padding-bottom: 2px;
-    letter-spacing: .5px;
   }
 
-  img {
-    margin-left: 10px;
+  &__account {
+    display: inline-block;
```
mmvsha: No need to make it `inline-block`, it's already a `flex-item`

src/components/Header/header.less
```diff
@@ -107,18 +108,34 @@
...
+
+    &--text {
+      line-height: 40px;
+      font-weight: 500;
+      padding-right: 5px;
```
mmvsha: why paddings instead of margins?

src/components/Header/index.js
```diff
@@ -39,22 +41,27 @@ const Header = ({ version, defaultAccount, currentBalance, currentProvider, getP
...
 
 Header.propTypes = {
   version: PropTypes.string,
-  defaultAccount: PropTypes.string,
+  currentNetwork: PropTypes.string,
+  currentAccount: PropTypes.string,
   currentBalance: PropTypes.string,
-  currentProvider: PropTypes.string,
-  getProviderIcon: PropTypes.func,
+  currentProvider: PropTypes.shape({}),
```
mmvsha: I think it will be good if you describe the shape at `src/utils/shapes.js`

src/components/ProviderIcon/index.js
```diff
@@ -0,0 +1,27 @@
...
+  [WALLET_PROVIDER.METAMASK]: 'metamask',
+  [WALLET_PROVIDER.PARITY]: 'parity',
+}
+
+const ProviderIcon = ({ provider = {} }) => (
+  <div
+    className={cn([
+      'headerIcon',
+      `headerIcon--${providerIconClasses[provider.name] || 'default'}`,
+      'pull-left',
```
mmvsha: No need in `pull-left` class name, it's a flex-item inside flexbox container

src/components/WalletIntegrationProvider/index.js
```diff
@@ -1,54 +1,59 @@
...
-import { getSelectedProvider } from 'selectors/blockchain'
-import { WALLET_PROVIDER } from 'integrations/constants'
-import Web3 from 'web3'
+import autobind from 'autobind-decorator'
+import { map } from 'lodash'
+import {
+  registerProvider,
+  updateProvider,
+  initGnosis,
+  setActiveProvider,
```
mmvsha: 'setActiveProvider' is defined but never used.

src/containers/App/index.js
```diff
@@ -61,16 +61,15 @@ class App extends Component {
...
   blockchainConnection: PropTypes.bool,
   children: PropTypes.node,
-  connectBlockchain: PropTypes.func,
   location: PropTypes.object,
   hasWallet: PropTypes.bool,
 }
 
 const mapStateToProps = state => ({
-  account: state.blockchain.defaultAccount,
+  activeProvider: state.blockchain.activeProvider,
```
mmvsha: Why is in other components  `getSelectedProvider` selector used and in this one a direct state access?

src/containers/HeaderContainer/index.js
```diff
@@ -1,32 +1,13 @@
...
-  return {
-    defaultAccount: getDefaultAccount(state),
-    currentBalance: getCurrentBalance(state),
-    currentProvider: currentProvider ? currentProvider.name : currentProvider,
-    getProviderIcon,
-  }
-}
+import { getCurrentAccount, getCurrentBalance, getCurrentNetwork, getSelectedProvider } from 'selectors/blockchain'
+
+const mapStateToProps = (state) => ({
```
mmvsha: [eslint] Unexpected parentheses around single function argument having a body with no curly braces (arrow-parens)

mmv08 (approved):

src/actions/blockchain.js
andre-meyer: I didn't create this var in this PR. I think 10 Seconds until a timeout is okay, 2 seconds on dev because we usually connect to localhost

mmv08 (approved):
Good job!

src/components/Header/header.less
```diff
@@ -14,6 +14,31 @@
...
+  &__network--text {
+    line-height: 40px;
+    font-weight: 300;
+    margin-right: 5px;
+    margin-bottom: 2px;
+    letter-spacing: .5px;
+  }
+
+  &__account {
+    display: inline-block;
```
mmvsha: No need in `display: inline-block` here, it's a flex item

mmv08 (approved):

PR #86, opened 2017-10-11
Task/time interval function fetching market

mmvsha:
**DESCRIPTION**:
This PR adds fetching market in background every 15 seconds. Previously after market creationg i had to refresh the page to see 'buy shares' button. Now it updates every 15 seconds and 'buy shares' button shows up if market has been mined.

**Tests**: 
1. Create market
2. Stay on market details page.
3. Wait a bit for `Buy shares` button to show up


src/api/rest.js
```diff
@@ -11,8 +11,9 @@ import {
 const API_URL = `${process.env.GNOSISDB_URL}/api`
 
 export const requestMarket = async marketAddress =>
-  restFetch(`${API_URL}/markets/${hexWithoutPrefix(marketAddress)}/`)
-    .then(response => normalize(response, marketSchema))
+  restFetch(`${API_URL}/markets/${hexWithoutPrefix(marketAddress)}/`).then(response =>
+    normalize({ ...response, local: false }, marketSchema),
```
apanizo: `local: false`? What is it?

src/components/MarketDetail/index.js
```diff
@@ -104,7 +104,25 @@ class MarketDetail extends Component {
...
   componentWillMount() {
+    if (!this.props.params.view) {
+      this.props.changeUrl(`/markets/${this.props.params.id}/${DEFAULT_VIEW}`)
+    }
+
+    this.fetchEssentialData()
+    this.props.requestGasCost(GAS_COST.BUY_SHARES)
+    this.props.requestGasCost(GAS_COST.SELL_SHARES)
+
+    this._fetchDataTimer = setInterval(this.fetchEssentialData, 15000)
```
apanizo: 15 seconds? @giacomolicari is that ok?

src/components/MarketDetail/index.js
```diff
@@ -104,7 +104,25 @@ class MarketDetail extends Component {
       marketFetchError: undefined,
     }
   }
+
   componentWillMount() {
+    if (!this.props.params.view) {
+      this.props.changeUrl(`/markets/${this.props.params.id}/${DEFAULT_VIEW}`)
```
apanizo: does the `changeUrl` stops the execution?

src/components/MarketDetail/index.js
giacomolicari: This can be parameterized inside config.json:
- fetchMarketTimeInterval: 150000

src/api/rest.js
mmvsha: @apanizo When we create a market it has property `local`, this means that market is saved locally in redux store and not in gnosisDB. We need some time to wait for gnosisDB to detect it. When it successfully fetched, we change `local` to false

src/components/MarketDetail/index.js
apanizo: In that case, would it make sense to introduce a `return`? or the lines 114-118 should be executed independently we go to another route?

src/components/MarketDetail/index.js
```diff
@@ -104,7 +105,25 @@ class MarketDetail extends Component {
...
+    }
+
+    this.fetchEssentialData()
+    this.props.requestGasCost(GAS_COST.BUY_SHARES)
+    this.props.requestGasCost(GAS_COST.SELL_SHARES)
+
+    this._fetchDataTimer = setInterval(this.fetchEssentialData, config.fetchMarketTimeInterval)
```
andre-meyer: This is a linter error, no? [`no-underscore-dangle`](https://eslint.org/docs/rules/no-underscore-dangle)

src/components/MarketDetail/index.js
```diff
@@ -104,7 +105,25 @@ class MarketDetail extends Component {
       marketFetchError: undefined,
     }
   }
+
   componentWillMount() {
+    if (!this.props.params.view) {
+      this.props.changeUrl(`/markets/${this.props.params.id}/${DEFAULT_VIEW}`)
```
andre-meyer: What is `DEFAULT_VIEW` here? Can you implement it so that it selects the first available tab? e.g. if you can't buy shares (the default view) it should show your current shares.

src/components/MarketDetail/index.js
mmvsha: @apanizo this is not an another route, this is default expandable tab on the same page

src/components/MarketDetail/index.js
mmvsha: @andre-meyer I haven't changed this part, it was there before :) Okay, i will try

src/components/MarketDetail/index.js
mmvsha: @andre-meyer Yep, it's there. I wasn't sure should I use underscore or no, but then I saw that its used in another files, so I used it. I'm not sure if we should add new eslint rule or not to use underscore

src/components/MarketDetail/index.js
andre-meyer: Oh! Okay, if it's not too much effort I mean. I just remember it from the QA talk and I thought you changed something there anyway :)

src/components/MarketDetail/index.js
andre-meyer: Really? I'd like to avoid changing the linter rules and just avoid underscores for now

PR #100, opened 2017-10-19
read only mode

mmvsha:


src/components/MarketList/index.js
```diff
@@ -93,13 +91,11 @@ class MarketList extends Component {
...
-          !isResolved && (
-            <div className="market__control">
-              <a href={`/#${resolveUrl}`} onClick={e => this.handleViewMarketResolve(e, resolveUrl)}>
-                Resolve
-              </a>
-            </div>
-          )}
+            !isResolved && (
+              <div className="market__control">
+                <a onClick={e => this.handleViewMarketResolve(e, resolveUrl)}>Resolve</a>
```
andre-meyer: Don't we need a href for `<a>` to work semantically?

src/components/MarketList/index.js
```diff
@@ -236,15 +232,26 @@ class MarketList extends Component {
...
               <div className="col-xs-10 col-xs-offset-1 col-sm-12 col-sm-offset-0">
-                <button
-                  type="button"
-                  onClick={this.handleCreateMarket}
-                  className="marketStats__control btn btn-default"
-                  disabled={!this.props.defaultAccount}
-                  title={this.props.defaultAccount ? '' : 'Please connect to an ethereum network to create a market'}
+                <Tooltip
+                  overlay="You need a wallet connected in order to create a market"
+                  trigger={this.props.defaultAccount ? [] : ['hover']}
```
andre-meyer: Can't you remove the tooltip completely if we don't have an account?

src/components/MarketList/index.js
```diff
@@ -236,15 +232,26 @@ class MarketList extends Component {
...
                 >
-                  Create Market
-                </button>
+                  <div
+                    className={cn({
+                      'marketStats__control--container': true,
+                      disabled: !this.props.defaultAccount,
```
andre-meyer: Can we have a selector for this and change the name? Something like `hasWallet`

src/components/Transactions/index.js
```diff
@@ -80,42 +77,58 @@ const renderTransaction = type => ({
...
+    )
+  }
+}
 
 Transactions.propTypes = {
-  runningTransactions: PropTypes.arrayOf(PropTypes.object),
+  changeUrl: PropTypes.func,
   completedTransactions: PropTypes.arrayOf(PropTypes.object),
+  currentAccount: PropTypes.string,
+  runningTransactions: PropTypes.arrayOf(PropTypes.object),
```
andre-meyer: Can you make a transaction Propshape? (not super important rn)

src/components/MarketList/index.js
mmvsha: @andre-meyer hmm, there is an eslint warning about this:
`Visible, non-interactive elements should not have mouse or keyboard event listeners (jsx-a11y/no-static-element-interactions)`
Probably we need to change it

src/components/MarketList/index.js
mmvsha: @andre-meyer Why? It tells user that he needs to connect a wallet

src/components/MarketList/index.js
mmvsha: @andre-meyer Selector for what? We have selector for current account: ```defaultAccount: getCurrentAccount(state),```
Shall i just rename the prop?

andre-meyer: Make a new selector that figures out if we have a wallet. Right now it should simply return false if we have no current account, then rename the prop

src/components/MarketList/index.js
andre-meyer: But you set the trigger to []?

andre-meyer: Oh! I meant if the user does have an account. We don't need the tooltip right?

src/components/MarketList/index.js
andre-meyer: how about the good old `href="javascript:void(0)"`

src/components/MarketList/index.js
```diff
@@ -204,6 +205,30 @@ class MarketList extends Component {
...
+        className={cn({
+          'marketStats__control--container': true,
+          disabled: !this.props.hasWallet,
+        })}
+      >
+        <button
+          type="button"
+          onClick={this.props.hasWallet ? this.handleCreateMarket : false}
+          className="marketStats__control btn btn-default"
+          disabled={!this.props.defaultAccount}
```
andre-meyer: use hasWallet here too

src/components/Header/index.js
```diff
@@ -9,21 +9,31 @@ import { upperFirst } from 'lodash'
...
+  currentBalance,
+  currentProvider,
+  getProviderIcon,
+  openConnectWalletModal,
+}) => (
   <div className="headerContainer">
     <div className="container">
       <div className="headerContainer__group headerContainer__group--logo">
-        <Link to="/">
+        <Link to={currentAccount ? '/' : '/markets/list'}>
```
andre-meyer: can you use the hasWallet selector here too?

src/components/Header/index.js
```diff
@@ -9,21 +9,31 @@ import { upperFirst } from 'lodash'
...
         </Link>
       </div>
       <div className="headerContainer__group headerContainer__group--left headerContainer__group--version">
         {version}
       </div>
       <div className="headerContainer__group headerContainer__group--left">
-        <Link to="/dashboard" activeClassName="headerContainer__navLink--active" className="headerContainer__navLink">
-          Dashboard
-        </Link>
+        {currentAccount && (
```
andre-meyer: Here too?

src/components/MarketDetail/index.js
```diff
@@ -145,8 +142,10 @@ class MarketDetail extends Component {
         })
       })
 
-    this.props.requestGasCost(GAS_COST.BUY_SHARES)
-    this.props.requestGasCost(GAS_COST.SELL_SHARES)
+    if (this.props.defaultAccount) {
```
andre-meyer: Use `hasWallet` selector

src/components/Header/index.js
```diff
@@ -41,16 +51,31 @@ const Header = ({ version, currentAccount, currentBalance, currentNetwork, curre
...
-        {currentAccount && currentProvider && (
-          <div className="headerContainer__account">
-            {currentNetwork && currentNetwork !== 'MAIN' && <span className="headerContainer__network--text">Network: {upperFirst(currentNetwork.toLowerCase())}</span>}
-            <DecimalValue value={currentBalance} className="headerContainer__account--text" />&nbsp;<span className="headerContainer__account--text">ETH</span>
-            <Identicon account={currentAccount} />
-          </div>
-        )}
-        {currentAccount && currentProvider && (
-          <ProviderIcon provider={currentProvider} />
+        {hasWallet && currentProvider && getProviderIcon(currentProvider)}
```
andre-meyer: Why `getProviderIcon` here? Is this not an image path?

PR #102, opened 2017-10-19
remove unused npm packages

mmvsha:


package.json
```diff
@@ -40,29 +42,20 @@
...
     "react-dom": "15.6.1",
-    "react-draggable": "3.0.3",
-    "react-intl": "2.4.0",
     "react-modal": "^2.4.1",
-    "react-portal": "^3.0.0",
     "react-redux": "^4.4.6",
     "react-router": "3.0.2",
     "react-router-redux": "4.0.7",
     "react-select": "^1.0.0-rc.10",
-    "react-textarea-autosize": "^4.0.5",
```
andre-meyer: I think this is used for the description of a market

mmvsha: @andre-meyer there is `react-autosize-textarea` which is actually used

PR #103, opened 2017-10-19
sell all the share amount if diff between wanted and share balance is…

mmvsha:
… less than 0.001

### DIFF
--- a/src/components/MarketMySharesForm/index.js
+++ b/src/components/MarketMySharesForm/index.js
@@ -11,7 +11,7 @@ import CurrencyName from 'components/CurrencyName'
 
 import FormInput from 'components/FormInput'
 
-import { COLOR_SCHEME_DEFAULT, GAS_COST, LOWEST_DISPLAYED_VALUE } from 'utils/constants'
+import { COLOR_SCHEME_DEFAULT, GAS_COST, LOWEST_DISPLAYED_VALUE, MIN_CONSIDER_VALUE } from 'utils/constants'
 import { getOutcomeName, weiToEth, normalizeScalarPoint } from 'utils/helpers'
 import { marketShape } from 'utils/shapes'
 
@@ -61,7 +61,7 @@ class MarketMySharesForm extends Component {
       const share = marketShares.filter(_share => _share.id === extendedSellId)[0]
       const fullAmount = Decimal(share.balance)
         .div(1e18)
-        .toDP(4, 1)
+        .toDP(2, 1)
         .toString()
       initialize({ sellAmount: fullAmount })
     }
@@ -87,7 +87,15 @@ class MarketMySharesForm extends Component {
   @autobind
   handleSellShare(shareId, shareAmount) {
     const shareIndex = this.props.marketShares.map(share => share.id).indexOf(shareId)
-    return this.props.sellShares(this.props.market, shareIndex, shareAmount).then(() => this.props.reset())
+    const shareBalance = new Decimal(this.props.marketShares[shareIndex].balance).div(1e18)
+    const selectedSellAmount = new Decimal(shareAmount)
+    const sellAmount = new Decimal(shareBalance)
+      .div(1e18)
+      .sub(selectedSellAmount)
+      .lt(MIN_CONSIDER_VALUE)
+      ? shareBalance
+      : shareAmount
+    return this.props.sellShares(this.props.market, shareIndex, sellAmount).then(() => this.props.reset())
   }
 
   @autobind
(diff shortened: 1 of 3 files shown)


src/components/MarketMySharesForm/index.js
```diff
@@ -87,7 +87,15 @@ class MarketMySharesForm extends Component {
   @autobind
   handleSellShare(shareId, shareAmount) {
     const shareIndex = this.props.marketShares.map(share => share.id).indexOf(shareId)
-    return this.props.sellShares(this.props.market, shareIndex, shareAmount).then(() => this.props.reset())
+    const shareBalance = new Decimal(this.props.marketShares[shareIndex].balance).div(1e18)
+    const selectedSellAmount = new Decimal(shareAmount)
+    const sellAmount = new Decimal(shareBalance)
+      .div(1e18)
+      .sub(selectedSellAmount)
```
andre-meyer: I think this value is not in Wei. If it is, can you change the variable name to indicate it's in wei? I think the value passed to handleSellShare is in ETH on Line 306

src/components/MarketMySharesForm/index.js
```diff
@@ -61,7 +56,7 @@ class MarketMySharesForm extends Component {
       const share = marketShares.filter(_share => _share.id === extendedSellId)[0]
       const fullAmount = Decimal(share.balance)
         .div(1e18)
-        .toDP(4, 1)
+        .toDP(2, 1)
```
andre-meyer: Why was this changed to 2?

src/components/MarketMySharesForm/index.js
```diff
@@ -87,7 +82,10 @@ class MarketMySharesForm extends Component {
   @autobind
   handleSellShare(shareId, shareAmount) {
     const shareIndex = this.props.marketShares.map(share => share.id).indexOf(shareId)
-    return this.props.sellShares(this.props.market, shareIndex, shareAmount).then(() => this.props.reset())
+    const shareBalance = new Decimal(this.props.marketShares[shareIndex].balance).div(1e18).toDP(2, 1)
+    const selectedSellAmount = new Decimal(shareAmount)
+    const sellAmount = shareBalance.sub(selectedSellAmount).lt(MIN_CONSIDER_VALUE) ? shareBalance : shareAmount
+    return this.props.sellShares(this.props.market, shareIndex, sellAmount).then(() => this.props.reset())
```
andre-meyer: can we use `async` + `await` here instead of promises?

src/components/MarketMySharesForm/index.js
```diff
@@ -87,7 +82,10 @@ class MarketMySharesForm extends Component {
   @autobind
   handleSellShare(shareId, shareAmount) {
     const shareIndex = this.props.marketShares.map(share => share.id).indexOf(shareId)
-    return this.props.sellShares(this.props.market, shareIndex, shareAmount).then(() => this.props.reset())
+    const shareBalance = new Decimal(this.props.marketShares[shareIndex].balance).div(1e18).toDP(2, 1)
+    const selectedSellAmount = new Decimal(shareAmount)
```
andre-meyer: This in Wei now, yes? ^^

mmvsha: @andre-meyer share.balance is in Wei, shareAmount is in ether

src/components/MarketMySharesForm/index.js
mmvsha: @andre-meyer because we are showing only 2 decimals for `share.balance` in the shares table, but initial value for input is 4 decimals

mmvsha: So it was like `5.50` in table, but the initial value for input was, for example, `5.5064`

src/components/MarketMySharesForm/index.js
mmvsha: @andre-meyer it was using promises, i didn't change anytyhing. I will then :smiley:

andre-meyer: Thank you! 👍

src/components/MarketMySharesForm/index.js
andre-meyer: But then it does `shareBalance (in wei) - shareAmountToSell (in ether)`, right? I think I'm confused

src/components/MarketMySharesForm/index.js
andre-meyer: Ok, got it

src/components/MarketMySharesForm/index.js
mmvsha: @andre-meyer `const shareBalance = new Decimal(this.props.marketShares[shareIndex].balance).div(1e18)`
It is divided before subtraction

PR #107, opened 2017-10-23
Feature/explain transactions

mmvsha:
This PR introduces transaction explanations and also has some ESLint fixes and precommit hook fix

- [x] Buy shares
- [x] Sell shares
- [x] Redeem winnings


src/reducers/modal.js
```diff
@@ -22,6 +22,7 @@ const reducer = handleActions({
 }, {
   isOpen: false,
   currentModal: undefined,
+  transactions: [],
```
andre-meyer: Can we turn this into a more generic "payload" attribute? I think we'll have more of these parameters we have to give to our modals. What do you think?

PR #109, opened 2017-10-25
fetch only whitelisted markets

mmvsha:


src/api/rest.js
```diff
@@ -10,13 +10,17 @@ import {
 
 const API_URL = `${process.env.GNOSISDB_URL}/api`
 
+const whitelistedAddressesFilter = Object.keys(process.env.WHITELIST)
+  .map((address, index) => `${index ? '&' : '?'}creator=${address}`)
+  .join('')
+
```
andre-meyer: Can we use a package for this? like `querystring`?

mmvsha: @andre-meyer yes, i will update it

mmvsha: But I dont think that it is really needed because for me it is simple enough without any packages

andre-meyer: It took me a while to understand that this generates a querystring and I think it would be cleaner to use a package to generate querystrings. We'll surely need querystrings again in the future and writing it yourself is errorprone with more complex objects

PR #116, opened 2017-10-27
change unit for scalar markets sell view

mmvsha:


src/components/MarketMySharesForm/index.js
```diff
@@ -316,7 +316,7 @@ class MarketMySharesForm extends Component {
                 <label>New predicted value</label>
                 <span>
                   <DecimalValue value={newScalarPredictedValue} />&nbsp;
-                  <CurrencyName collateralToken={market.event.collateralToken} />
+                  <CurrencyName outcomeToken={market.eventDescription.unit} />
```
andre-meyer: The scalar outcome unit is not the currency, this component is used for displaying ERC20 tokens like the "Short" token or the "Long" token. Please just output `{market.eventDescription.unit}` here. In the future we can pass the contract address here and the "CurrencyName" component will resolve the name

src/components/CurrencyName/index.js
```diff
@@ -2,7 +2,7 @@ import React from 'react'
 import PropTypes from 'prop-types'
 
 // Current mapping does not contain any logic
-export const outcomeTokenToText = () => 'Outcome Token'
+export const outcomeTokenToText = outcomeToken => outcomeToken
```
andre-meyer: Please also change this back

PR #134, opened 2017-11-08
Backmerge #1 from Olympia Interface

andre-meyer:


apanizo (approved):
Just that? I thought it would be worse...

PR #139, opened 2017-11-10
Feature PM-19 have the markets shown a bit up in the page

mmvsha:


src/components/MarketList/index.js
```diff
@@ -94,7 +94,10 @@ class MarketList extends Component {
           {isOwner &&
             !isResolved && (
               <div className="market__control">
-                <a href="javascript:void(0)" onClick={e => this.handleViewMarketResolve(e, `/markets/${market.address}/resolve`)}>
+                <a
```
apanizo: This method should be placed in a separate constant otherwise always this component will be re-rendered even if it not necessary.

src/components/MarketList/index.js
```diff
@@ -231,21 +230,23 @@ class MarketList extends Component {
...
-                  className="marketStats__control btn btn-default"
-                  whitelistRequired
-                >
-                  Create Market
-                </InteractionButton>
+        {process.env.WHITELIST[defaultAccount] && (
+          <div className="marketListPage__controls">
+            <div className="container">
+              <div className="row">
+                <div className="col-xs-10 col-xs-offset-1 col-sm-12 col-sm-offset-0">
```
apanizo: I think this should be ```col-xs-10 col-xs-offset-2 col-sm-12``` as far as I know the grid layout we are using is divided in 12 columns, right?

PR #145, opened 2017-11-16
Transactions page refactoring

mmvsha:


src/components/ProgressIndicator/index.js
```diff
@@ -0,0 +1,34 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import ProgressSpinner from 'components/ProgressSpinner'
+import { TRANSACTION_COMPLETE_STATUS } from 'utils/constants'
+
+const ProgressIndicator = ({ completed, completionStatus, progress }) => {
```
apanizo: As you are creating a new component due to a refactor I suggest:
```
  const iconType = completionStatus === TRANSACTION_COMPLETE_STATUS.NO_ERROR ? 'checkmark' : 'error'
  return (
      <Block>
          { completed
              ? <div className="transaction__icon">  <--- This should be Block as well but let's leave div for simplicity
                  <div className="`icon icon--${iconType}`" />
              </div>
              : <ProgressSpinner
                  width={32}
                  height={32}
                  strokeWidthPx={1}
                  fontSizePx={8}
                  progress={progress}
                  modifier="spinning"
               />
      </Block>
  )
```

src/components/Transaction/index.js
```diff
@@ -0,0 +1,57 @@
...
+  [TRANSACTION_COMPLETE_STATUS.LOST]: 'Because of a page reload, we lost the status of this Transaction',
+}
+
+const Transaction = ({ label, startTime, endTime, completed, completionStatus, progress, type }) => (
+  <div className={`transactionsPage__transaction transactionsPage__transaction--${type} transaction`}>
+    <ProgressIndicator completed={completed} completionStatus={completionStatus} progress={progress} />
+    <div className="transaction__content">
+      <div className="transaction__heading">{label}</div>
+      <div className="transaction__details">
+        <div className="transaction__detail">
```
apanizo: I would create every *transaction__detail* into a separate PRIVATE component (so place them as constants in this file). As far as I see there are three different:
* Created at
* Finished at
* Transaction time.

Suggestions:
I think those components could be reused/refactored into one, the differences between them are the text and the children props.

src/components/Transactions/index.js
```diff
@@ -119,7 +41,9 @@ class Transactions extends Component {
               There are no previous transactions
             </div>
           )}
-          {completedTransactions.map(renderTransaction('completed'))}
+          {completedTransactions.map(transaction => (
```
apanizo: The final *(* is necessary?

PR #147, opened 2017-11-16
Bug PM-183 Fix closed/resolved message show conditions for markets

mmvsha:
This PR includes fix for showing closed/resolved message condition, eslint/codestyle fixes for ScalarSlider component, PropType fixes with change of type of Marginal Prices

### DIFF
--- a/src/components/MarketDetail/index.js
+++ b/src/components/MarketDetail/index.js
@@ -248,6 +248,7 @@ class MarketDetail extends Component {
     const marketClosed = market.stage === MARKET_STAGES.MARKET_CLOSED
     const marketResolved = market.oracle.isOutcomeSet
     const marketClosedOrFinished = marketClosed || marketResolved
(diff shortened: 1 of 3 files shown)


src/components/MarketDetail/index.js
```diff
@@ -294,7 +293,9 @@ class MarketDetail extends Component {
                 .format(RESOLUTION_TIME.ABSOLUTE_FORMAT)}
             </div>
             {marketClosedOrFinished && (
-            <div className="marketTimer__marketClosed">{timeToResolution < 0 ? 'This market was resolved.' : 'This market was closed.'}</div>
+              <div className="marketTimer__marketClosed">
+                {`This market was ${marketResolved ? 'resolved.' : 'closed.'}`}
```
apanizo: I would move this to a constant ```const status = marketResolved ? 'resolved.' : 'closed.'``` and then
```{`This market was ${status}`}```

If tomorrow we have a different status we only have to change one variable.

PR #150, opened 2017-11-20
Bug: Fix Header Flex Styling

andre-meyer:
**Previous**
image

**After**
image

### DIFF
--- a/src/components/Header/header.less
+++ b/src/components/Header/header.less
@@ -65,6 +65,12 @@
   }
 
   &__account {
+    vertical-align: middle;
+    flex-direction: row;
+    flex-wrap: nowrap;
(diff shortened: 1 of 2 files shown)


apanizo:
@andre-meyer I like the before/after in description

src/components/Header/header.less
```diff
@@ -65,6 +65,12 @@
   }
 
   &__account {
+    vertical-align: middle;
+    flex-direction: row;
+    flex-wrap: nowrap;
+    justify-content: center;
+    align-items: center;
+
```
apanizo: Just one note, if there is no message about the incorrect network, does that css won't place the account info in the middle of the space after the logo/menu?

PR #151, opened 2017-11-20
Feature PM-53 Unity the terms open/ended

mmvsha:
This PR changes open/closed to open/ended, also adds calculations for markets overview

### DIFF
--- a/src/components/MarketList/index.js
+++ b/src/components/MarketList/index.js
@@ -30,11 +30,11 @@ const resolutionFilters = [
     value: '',
   },
   {
-    label: 'Resolved',
+    label: 'Open',
     value: 'RESOLVED',
   },
   {
-    label: 'Unresolved',
+    label: 'Ended',
     value: 'UNRESOLVED',
   },
 ]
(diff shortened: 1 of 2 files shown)


src/components/MarketList/index.js
```diff
@@ -202,6 +200,17 @@ class MarketList extends Component {
   render() {
     const { markets, defaultAccount } = this.props
 
+    const threeDayMSeconds = 3 * 24 * 60 * 60 * 1000
+    const openMarketsAmount = markets.filter(
+      ({ stage, oracle: { isOutcomeSet } }) => stage !== MARKET_STAGES.MARKET_CLOSED && !isOutcomeSet,
+    ).length
+    const endingSoonMarketsAmount = markets.filter(
+      ({ eventDescription: { resolutionDate } }) => new Date(resolutionDate) - new Date() < threeDayMSeconds,
```
apanizo: In general, I suggest moving as a constant the ``` new Date(...)``` objects. You are consuming memory resources for every iteration.

Seeing you are calculating stuff based on "3 days margin" you can safely put constants before the filters.

Something like 
```
const now = new Date()
```

mmvsha: @apanizo Ok, so this will be the date user opened the page, but what if user spends 5 minutes on the site and then comes back to see new markets. The constant wont change and we will show old information.

apanizo: If the component will be re-rendered, the ```const now = new Date()``` will be updated. Anyway as far as I see you are comparing against **threeDayMSeconds**, does it matter 5 minutes?

mmvsha: Ah, I thought you were talking about moving the constant out of component and now the filter method

apanizo: :) No I meant:
```
    render() {
        const now = new Date()
        ....
        const endingSoonMarketsAmount = markets.filter(
            ({ eventDescription: { resolutionDate } }) => new Date(resolutionDate) - now < threeDayMSeconds,
         ....
```

But, as I said I do not think it would be a big deal put ```now``` constant outside the render function because it is comparing dates with 3 days of difference :)

apanizo:
Please, resolve conflicts

PR #168, opened 2017-12-01
Bug PM-237 fix css for dashboard market list items / make market clickable

mmvsha:
I moved onClick event listener from market title to market container. Also I've changed CSS so long outcomes will show as "Long outc..." and dont break our mark up

**Before:** 
screen shot 2017-12-05 at 12 54 30

**After:**
screen shot 2017-12-05 at 12 53 24


src/components/Dashboard/dashboard.less
```diff
@@ -348,6 +358,13 @@
     color: #ffffff;
     margin-top: 6px;
     display: inline-block;
+
+    &.listItemOutcome {
+      white-space: nowrap;
+      overflow: hidden;
+      text-overflow: ellipsis;
+      width: 76%;
```
andre-meyer: Hmm, does this work with all screen sizes?

src/components/Dashboard/dashboard.less
```diff
@@ -348,6 +358,13 @@
     color: #ffffff;
     margin-top: 6px;
     display: inline-block;
+
+    &.listItemOutcome {
```
andre-meyer: Can we use BEM here? I know we decided to go for a different approach in the future, but for now we should stick to it:
`&__listItem`

src/components/Dashboard/dashboard.less
```diff
@@ -184,6 +184,16 @@
...
       align-items: center;
     }
+
+    .entry__color {
+      margin-bottom: 0;
+    }
+  }
+  
+  
+  .listItemOutcomeContainer {
```
andre-meyer: same here, BEM please

src/components/Dashboard/dashboard.less
apanizo: I disagree. The only reasons I can think of for sticking to an old approach are: we are going to change something in bulk or it is not super related and were are in a hurry/hotfix mode. 

Each day I have more doubts we will have time for refactoring so, seeing the project is "stabilized" it is perfectly fine to start/continue with the new CSS approach.

src/components/Dashboard/index.js
```diff
@@ -139,10 +139,12 @@ class Dashboard extends Component {
...
       if (market) {
         return (
-          <div className="dashboardMarket dashboardMarket--onDark" key={index}>
-            <div className="dashboardMarket__title" onClick={() => this.handleViewMarket(market)}>
-              {holding.eventDescription.title}
-            </div>
+          <div
+            className="dashboardMarket dashboardMarket--onDark"
+            key={index}
+            onClick={() => this.handleViewMarket(market)}
```
apanizo: Extract into a variable otherwise, this component will be re-rendered always.

src/components/Dashboard/dashboard.less
mmvsha: @andre-meyer It works the same as our interface, not ready for small screens

src/components/Dashboard/index.js
```diff
@@ -86,34 +86,37 @@ class Dashboard extends Component {
...
-        <Outcome market={market} opts={{ showOnlyTrendingOutcome: true, showDate: true, dateFormat: 'MMMM Y' }} />
-      </div>
-    ))
+    return markets.map((market) => {
+      const viewMarket = () => this.handleViewMarket(market)
+
+      return (
+        <div className="dashboardMarket dashboardMarket--new" key={market.address} onClick={viewMarket}>
+          <div className="dashboardMarket__title">{market.eventDescription.title}</div>
+          <Outcome market={market} opts={{ showOnlyTrendingOutcome: true, showDate: true, dateFormat: 'MMMM Y' }} />
```
apanizo: Objects should not be created here.

src/components/Dashboard/index.js
```diff
@@ -86,34 +86,37 @@ class Dashboard extends Component {
...
+          className="dashboardMarket dashboardMarket--closing dashboardMarket--twoColumns"
+          key={market.address}
+          onClick={viewMarket}
+        >
+          <div className="dashboardMarket__leftCol">
+            <div className="value">{moment.utc(market.eventDescription.resolutionDate).fromNow()}</div>
+          </div>
+          <div className="dashboardMarket__rightCol">
+            <div className="dashboardMarket__title">{market.eventDescription.title}</div>
+            <Outcome market={market} opts={{ showOnlyTrendingOutcome: true }} />
```
apanizo: And here.

PR #178, opened 2017-12-12
Fix redeemWinnings flow

cag:
If the contract exists, will exit the function as intended. Will also ensure WinningsRedemption event is fired.


cag:
I should have made the change on Olympia, as I've noticed that the lack of a `return` statement has already been addressed there: https://github.com/gnosis/olympia-interface/pull/120/

PR #184, opened 2017-12-19
Backmerge: Selector Refactors, Wallet Refactor/Integration Refactor, Bugfixes

andre-meyer:
PM-358
PM-295
PM-275
PM-300
PM-311
PM-333
Refactor: Trading/Graph Selector
PM-191

**Squash on Merge!!!**


andre-meyer:
Sorry for the huge PR

src/utils/analytics/gaSend.js
```diff
@@ -0,0 +1,5 @@
+import ga from './ga'
+
+export default (props) => {
+  ga('olympiatracker.send', ...props)
```
apanizo: Do you really want to integrate GA in m-i? If so I miss the AnalyticsId in ```init.js``` and custom events names

andre-meyer: you're right, I'll get rid of it for now

PR #195, opened 2017-12-22
Fix integrations watcher

mmvsha:
**BEFORE:**
Changing account, balance, locking metamask is not recognized in the interface

**AFTER:**
Changing account, balance, locking metamask is recognized in the interface

Actual changes are in:
- `src/integrations/metamask/index.js`
- `src/integrations/injectedWeb3.js`

Other files are only eslint fixes

### DIFF
--- a/src/integrations/metamask/index.js
+++ b/src/integrations/metamask/index.js
@@ -13,6 +13,15 @@ class Metamask extends InjectedWeb3 {
    * This allows "fallback providers" like a remote etherium host to be used as a last resort.
    */
   static providerPriority = 90
+  static watcherInterval = 1000
+
+  constructor() {
+    super()
+    this.watcher = setInterval(() => {
+      this.watch('account', this.getAccount)
+      this.watch('balance', this.getBalance)
+    }, Metamask.watcherInterval)
+  }
 
   /**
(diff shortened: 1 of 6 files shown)


src/integrations/metamask/index.js
```diff
@@ -27,7 +36,10 @@ class Metamask extends InjectedWeb3 {
     this.walletEnabled = false
 
     try {
-      if (typeof window.web3 !== 'undefined' && window.web3.currentProvider.constructor.name === 'MetamaskInpageProvider') {
+      if (
+        typeof window.web3 !== 'undefined' &&
```
apanizo: Why do we have this condition here? What does it do?

src/integrations/metamask/index.js
```diff
@@ -13,6 +13,15 @@ class Metamask extends InjectedWeb3 {
...
    */
   static providerPriority = 90
+  static watcherInterval = 1000
+
+  constructor() {
+    super()
+    this.watcher = setInterval(() => {
+      this.watch('account', this.getAccount)
+      this.watch('balance', this.getBalance)
+    }, Metamask.watcherInterval)
```
apanizo: Only account and balance? What about network and networkId?

PR #205, opened 2018-01-04
Make reflowable column inline-block to shrinkwrap

cag:
This is so clicks on market list items are not blocked on narrow width displays

Just thought I would start up the [Cipher support](https://www.cipherbrowser.com/)


cag:
Sure!

Before:
screenshot from 2018-01-04 16-25-25
screenshot from 2018-01-04 16-26-24

After:
screenshot from 2018-01-04 16-27-26
screenshot from 2018-01-04 16-27-16

PR #206, opened 2018-01-09
PM-229: Fix "Withdraw fees" button not triggering the transaction

mmvsha:
**BEFORE**:
After pressing `Withdraw fees` transaction is not triggered.

**AFTER**:
After pressing `Withdraw fees` transaction is triggered.

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,7 +32,7 @@
     "precommit"
   ],
   "dependencies": {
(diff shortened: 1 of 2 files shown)


mmvsha:
@apanizo It's the same in Olympia, so I assume no, it isn't

PR #210, opened 2018-01-16
Backmerge: Projectstructure, Linters, Configuration, Package.json

andre-meyer:
### Backmerge (Redo) 01

Includes changes for:
- Entrypoint `index.js`
- Store Configuration `store.js`
- General Projectstructure
- Linting Configuration
- `package.json`
- Configuration Files


nginx.template
```diff
@@ -18,6 +18,14 @@ server {
 
   keepalive_timeout   70;
   add_header          Alternate-Protocol  $NGINX_SSL_PORT:npn-spdy/2;
+
+  location ~* ^.+.(jpg|jpeg|png|gif|ico|svg|css|js)$ {
+    expires 30d;
```
apanizo: What is this? 30d cache for js files?

package.json
```diff
@@ -48,6 +52,7 @@
     "normalizr": "^3.2.2",
     "prop-types": "^15.6.0",
     "raven-js": "^3.21.0",
+    "raven-for-redux": "^1.1.1",
```
apanizo: What does it do?

src/config.json
```diff
@@ -17,5 +20,6 @@
     "host": "rinkeby.infura.io",
     "port": 443
   },
-  "fetchMarketTimeInterval": 15000
+  "fetchMarketTimeInterval": 15000,
+  "ravenPublicDSN": null
```
apanizo: What is this?

webpack.dev.config.js
```diff
@@ -106,18 +102,25 @@ module.exports = {
...
     }),
-    new webpack.DefinePlugin({
-      'process.env': {
-        VERSION: JSON.stringify(`${version}#${build}`),
-        NODE_ENV: JSON.stringify(nodeEnv),
-        GNOSISDB_URL: JSON.stringify(gnosisDbUrl),
-        ETHEREUM_URL: JSON.stringify(ethereumUrl),
-        WHITELIST: JSON.stringify(whitelist),
-      },
+    new webpack.EnvironmentPlugin({
```
apanizo: Are you sure we want to use ```EnvironmentPlugin```? IF my memory does not fail we switched to DefinePlugin, right?

I specifically remember during the last deployment Giacomo told me: "Use ```WHITELIST: JSON.stringify(whitelist),```.

What am I missing?

webpack.prod.config.js
```diff
@@ -94,18 +93,25 @@ module.exports = {
...
-    new webpack.DefinePlugin({
-      'process.env': {
-        VERSION: JSON.stringify(`${version}#${build}`),
-        NODE_ENV: JSON.stringify(nodeEnv),
-        GNOSISDB_URL: JSON.stringify(gnosisDbUrl),
-        ETHEREUM_URL: JSON.stringify(ethereumUrl),
-        WHITELIST: JSON.strinfigy(whitelist),
-      },
+    new webpack.EnvironmentPlugin({
+      VERSION: `${version}#${build}`,
```
apanizo: Same here?

nginx.template
denisgranha: we don't use nginx anymore for our servers, so this file is not relevant

package.json
andre-meyer: it's for sentry, I'm already adding it

src/config.json
andre-meyer: Sentry.io, already added the key

webpack.dev.config.js
andre-meyer: Nope, we use `EnvironmentPlugin` on both `webpack.prod.js` and `webpack.dev.js`. Doesn't require to use stringify

webpack.prod.config.js
andre-meyer: Nope nope, Environment Plugin just like in Olympia

webpack.dev.config.js
apanizo: @denisgranha @giacomolicari ? Will we have any problem?

denisgranha: Don't remember the scenario but the whitelisted accounts weren't being used until we added stringify.
Test your solution with both enviroments and whatever is better
I'm not an expert about those plugins
It's up to you

nginx.template
andre-meyer: Will remove it

webpack.dev.config.js
andre-meyer: https://webpack.js.org/plugins/environment-plugin/#usage

This plugin automatically `stringify`s all values :)

andre-meyer: I had Mikhail test it for olympia, it's in both configurations now and the last problem we had, had nothing to do with this

apanizo (approved):

PR #211, opened 2018-01-17
Use Ganache-cli instead of TestRPC in README

Uxio0:
Default port has also been changed from 8000 to 5000

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,13 @@
 gnosis-management
 ------
 
-Start with `npm start`, server available under `http://localhost:8000/`
+Start with:
+```
+npm install
+npm start
+```
+
+Server will be available under `http://localhost:5000/`
 
 Installing gnosis.js
 ------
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -28,6 +34,6 @@ Quick version:
 - Build the image and start the container with `docker-compose build && docker-compose up`.
 - Enter the docker container with `docker-compose run web bash`.
 - Inside the docker container run `python manage.py createsuperuser` and create a user.
-- Login at `localhost:8000/admin` with the created user.
+- Login at `localhost:5000/admin` with the created user.
```
andre-meyer: I think this is for GnosisDB Django Admin. Can you check?

Uxio0: You are right Andre, I'm pushing the fix

andre-meyer:
Thanks @Uxio0  👍

PR #212, opened 2018-01-17
Backmerge: Dashboard Page

andre-meyer:
### Backmerge (Redo) 02

Includes changes for:
- `DashboardPage` Components and Container
- `market.js` Selector and Action


src/containers/DashboardPage/index.js
```diff
@@ -1,23 +1,15 @@
...
-import { getCurrentAccount, getEtherTokensAmount, isGnosisInitialized, checkWalletConnection } from 'selectors/blockchain'
+import { getMarkets, getAccountPredictiveAssets } from 'selectors/market'
+import { getAccountTrades } from 'selectors/marketTrades'
+import { getAccountShares } from 'selectors/marketShares'
+import { getCurrentAccount, getEtherTokensAmount, isGnosisInitialized } from 'selectors/blockchain'
 import { requestMarkets, requestAccountTrades, requestAccountShares, redeemWinnings } from 'actions/market'
 import { requestGasPrice, requestEtherTokens } from 'actions/blockchain'
 import { weiToEth } from 'utils/helpers'
-
+import { areCredentialsValid } from 'integrations/uport/connector'
```
mmvsha: Do we need it?

src/actions/market.js
```diff
@@ -123,8 +119,8 @@ export const requestMarketTradesForAccount = (marketAddress, accountAddress) =>
  * Requests all trades (tradehistory) on a market from GnosisDB.
  * @param {Market} market
  */
-export const requestMarketTrades = marketAddress => async (dispatch) => {
-  const payload = await api.requestMarketTrades(marketAddress)
+export const requestMarketTrades = market => async (dispatch) => {
```
mmvsha: https://github.com/gnosis/olympia-interface/pull/173/files#diff-a7e638fe149ee1e4fb43773cc5f14672R122
This change is from my PR which is still in testing 😮

andre-meyer: It's okay, I talked to Lama, it works 👍

src/actions/market.js
```diff
@@ -377,6 +382,7 @@ export const sellMarketShares = (market, share, outcomeTokenCount, earnings) =>
 
   // Reset the allowance if the cost of current transaction is greater than the current allowance
```
mmvsha: These comments (also next line) are out of place, you can check how it's done in Olympia: https://github.com/gnosis/olympia-interface/blob/master/src/actions/market.js#L279

src/actions/market.js
```diff
@@ -493,6 +507,8 @@ export const withdrawFees = market => async (dispatch) => {
     throw e
   }
 
+  dispatch(refreshTokenBalance())
+
   // TODO: Update market so we can't withdraw again
```
mmvsha: It has been already done so the comment can be removed

src/components/Dashboard/index.js
```diff
@@ -207,23 +176,22 @@ class Dashboard extends Component {
       const viewMarket = () => this.handleViewMarket(market)
 
       return (
-        <div className="dashboardMarket dashboardMarket--onDark" key={index} onClick={viewMarket}>
-          <div className="dashboardMarket__title">{trade.eventDescription.title}</div>
+        <div
+          className="dashboardMarket dashboardMarket--onDark"
+          key={index}
+          onClick={() => this.handleViewMarket(market)}
```
mmvsha: You can use `viewMarket` function from line 176 here

mmv08 (changes requested):

src/actions/market.js
```diff
@@ -5,22 +5,18 @@ import uuid from 'uuid/v4'
 import * as api from 'api'
 
 import { receiveEntities, updateEntity } from 'actions/entities'
-
+import { openModal, closeModal } from 'actions/modal'
+import { refreshTokenBalance } from 'actions/blockchain'
```
apanizo: There is something that I do not understand. If you take a look into the ```refreshTokenBalance```  function, you will notice a couple of things. 

First I was not able to find in ```src/actions/blockchain``` in ```development``` or ```backmerge/redo```  branches.  

The second, that function seems to be created for Olympia purposes, do not know if that function should be in management or not. 

My guess is we have to remove it because we are getting the balance from the provider and we do not have a collateral token.

src/components/Dashboard/index.js
```diff
@@ -108,31 +107,6 @@ class Dashboard extends Component {
...
-          <div className="dashboardMarket__rightCol">
-            <div className="dashboardMarket__title">{market.eventDescription.title}</div>
-            <Outcome market={market} opts={outcomeOptions} />
-          </div>
-        </div>
-      )
-    })
-  }
-
   renderMyHoldings(holdings) {
```
apanizo: Do not know if refactor should be done at this point, but I would extract MyHoldings to a private component and the reduces the complexity of lines 132-138

src/components/Dashboard/index.js
```diff
@@ -234,11 +202,15 @@ class Dashboard extends Component {
...
 
   renderWidget(marketType) {
-    const {
-      markets, accountShares, accountTrades,
-    } = this.props
-
-    const whitelistedMarkets = markets.filter(market => process.env.WHITELIST[market.creator] && !isMarketResolved(market) && !isMarketClosed(market))
+    const { markets, accountShares, accountTrades } = this.props
+
+    const whitelistedMarkets = markets.filter(market =>
```
apanizo: This is complex. That logic should be moved into a local selector (in selector.js file) and just receive valid markets on the component.

apanizo (changes requested):

src/actions/market.js
andre-meyer: 👍

src/components/Dashboard/index.js
apanizo: Is this going to be implemented now?

andre-meyer: Please let's do it later when we touch this file for the refactor

apanizo: Ok, sounds good.

PR #214, opened 2018-01-24
Backmerge: Market Details

mmvsha:
Includes changes for:
- API
- utils
- bunch of components used on Market Detail page
- selectors


src/api/gnosis.js
```diff
@@ -139,7 +138,7 @@ export const createEvent = async (event) => {
 
   const eventData = {
     ...event,
-    collateralToken: hexWithPrefix(gnosis.etherToken.address),
+    collateralToken: hexWithPrefix(gnosis.olympiaToken.address),
```
mmvsha: This is related to market creation and since we are going to remove it, I decided not to fix that 🙂

src/api/gnosis.js
```diff
@@ -155,14 +154,17 @@ export const createEvent = async (event) => {
     creator: await getCurrentAccount(),
     creationDate: moment().format(),
     ...event,
-    collateralToken: hexWithPrefix(gnosis.etherToken.address),
+    collateralToken: hexWithPrefix(gnosis.olympiaToken.address),
```
mmvsha: same here

src/api/gnosis.js
```diff
@@ -204,7 +206,7 @@ export const fundMarket = async (market) => {
     await gnosis.etherToken.deposit({ value: marketFundingWei.toString() })
   }
 
-  const marketAllowance = await gnosis.etherToken.allowance(
+  const marketAllowance = await gnosis.olympiaToken.allowance(
```
mmvsha: and here

src/components/CurrencyName/index.js
```diff
@@ -2,7 +2,7 @@ import React from 'react'
 import PropTypes from 'prop-types'
 
 // Current mapping does not contain any logic
-export const outcomeTokenToText = () => 'Outcome Token'
+export const outcomeTokenToText = outcomeToken => outcomeToken
```
apanizo: A function which returns the same as it receives as a parameter? Why do we need a function for that?

src/components/DecimalValue/index.js
```diff
@@ -21,17 +21,15 @@ export const decimalToText = (value, decimals = 2) => {
...
 
-const DecimalValue = ({ value, decimals = 2, className }) => {
+const DecimalValue = ({ value, decimals = 4, className }) => {
   const text = decimalToText(value, decimals)
   return <span className={className}>{text}</span>
 }
 
 // I don't use PropTypes.instanceOf because Decimal can be cloned with different default properties
 // and instanceOf doesn't deal with that situation. In fact, Decimal.clone is used in gnosis.js
 const decimalJsTest = (props, propName, componentName) => {
```
apanizo: Can you provide me with an example? And where and why do we use this function?

src/components/MarketDetail/index.js
```diff
@@ -294,6 +297,19 @@ class MarketDetail extends Component {
     )
   }
 
+  renderMarketGraph() {
+    const { market, marketGraph } = this.props
+    if (!marketGraph.length) {
```
apanizo: What about:
```
const hasData = marketGraph.length;
....

<div className="container">
    { hadData
        ? <MarketGraph data={marketGraph} market={market} />
        :  <LoadingIndicator className="marketGraph__spinner" />
    }
</div>
```

src/components/MarketMySharesForm/index.js
```diff
@@ -61,11 +67,11 @@ class MarketMySharesForm extends Component {
     const { extendedSellId } = this.state
     const { selectedSellAmount, marketShares, initialize } = this.props
     const sellAmountAndMarketSharesAreDefined =
-      selectedSellAmount === undefined && extendedSellId !== undefined && marketShares.length
+      selectedSellAmount === undefined && extendedSellId !== undefined && Object.keys(marketShares).length
```
apanizo: We are using in many places ```.length``` for calculating if variables have data. I think when we do the refactoring we should extract this function somewhere else.

src/components/MarketMySharesForm/index.js
```diff
@@ -89,9 +95,16 @@ class MarketMySharesForm extends Component {
     this.setState({ extendedSellId: shareId === this.state.extendedSellId ? undefined : shareId })
```
apanizo: If we are updating the state based on a previous value of itself, you have to do it with a function, otherwise you can not guarantee the state will be updated in order. Something like:
```
this.setState((state) => {extendedSellId : shareId === this.state.extendedSellId ? undefined : shareId})
```

src/components/MarketMySharesForm/index.js
```diff
@@ -143,58 +147,42 @@ class MarketMySharesForm extends Component {
...
         <td>
-          <div
-            className="shareOutcome__color"
-            style={{ backgroundColor: COLOR_SCHEME_DEFAULT[share.outcomeToken.index] }}
-          />
+          <div className="shareOutcome__color" style={outcomeColorStyle} />
         </td>
         <td className="">{getOutcomeName(market, share.outcomeToken.index)}</td>
         <td>
           {Decimal(share.balance)
```
apanizo: Render function should be simple, I suggest extract this condition to a variable and simplify this code.

src/components/MarketMySharesForm/index.js
```diff
@@ -143,58 +147,42 @@ class MarketMySharesForm extends Component {
...
           {!resolvedOrClosed && (
-            <a
-              href="javascript:void(0);"
-              className="marketMyShares__sellButton"
-              onClick={e => this.handleShowSellView(e, share.id)}
-            >
+          <a
+            href="javascript:void(0);"
+            className="marketMyShares__sellButton"
+            onClick={e => this.handleShowSellView(e, share.id)}
```
apanizo: Arrow functions should be avoided in the render function, otherwise, you will re-render every time a prop or states or parent changes.

Also, try to change the ```<a>``` and use ```Link```, it should work I think.

src/components/MarketMyTrades/index.js
```diff
@@ -33,49 +34,50 @@ class MarketMyTrades extends Component {
...
-            '0'
-          )}
-        </td>
-      </tr>
-    ))
+    const tableRowElements = marketTrades.map((trade) => {
+      const outcomeColorStyle = { backgroundColor: colorScheme[trade.outcomeToken.index] }
+
+      return (
+        <tr className="marketMyTrades__share" key={trade.id}>
```
apanizo: This function will be refactored, right?

src/components/OutcomePriceChanged/index.js
```diff
@@ -5,7 +5,8 @@ import '../TransactionsExplanation/TransactionExplanation.less'
 const OutcomePriceChanged = ({ closeModal }) => (
   <div className="transactionsExplanation">
     <a className="transactionsExplanation__close" href="javascript:void(0);" onClick={() => closeModal()} />
```
apanizo: <a> and arrow function as a prop.

mmvsha:
@apanizo Fully agree with all your feedback, but let's do it after we finish backmerging everything

apanizo:
:( I have the sensation we are placing many things to the "refactor phase", hopefully, we won't forget anything.

It is alright. Let's go.

src/components/CurrencyName/index.js
andre-meyer: This was put in place for when we implement loading the outcome Tokens name from the Blockchain. It's still not needed, it could just as well be made a string. But not in the Backmerge

src/components/DecimalValue/index.js
andre-meyer: `new Decimal()` and `Decimal.clone` don't work with `PropTypes.instanceOf` for some reason. I can't remember where it was, but instanceOf never worked, so a custom proptype-test was put in place.

src/components/MarketMyTrades/index.js
```diff
@@ -4,11 +4,12 @@ import Decimal from 'decimal.js'
 import moment from 'moment'
 import { decimalToText } from 'components/DecimalValue'
 import CurrencyName from 'components/CurrencyName'
-import { COLOR_SCHEME_DEFAULT, RESOLUTION_TIME } from 'utils/constants'
+import { RESOLUTION_TIME } from 'utils/constants'
 import { getOutcomeName, weiToEth } from 'utils/helpers'
 import { marketShape, marketTradeShape } from 'utils/shapes'
 
 import './marketMyTrades.less'
+import { COLOR_SCHEME_SCALAR, COLOR_SCHEME_DEFAULT, OUTCOME_TYPES } from '../../utils/constants'
```
andre-meyer: Merge together with imports from Line 7

src/components/MarketMySharesForm/index.js
andre-meyer: I know we need to avoid this, but I think the reason is a different one. A prop or a state change will (and should) cause re-renders, but we're leaking memory by _recreating_ the function every render. I think also the performance impact on this is very very low.

Take a look at this article from the React Docs: https://reactjs.org/docs/faq-functions.html#is-it-ok-to-use-arrow-functions-in-render-methods

src/components/MarketDetail/index.js
mmvsha: @apanizo VM119:1 Uncaught ReferenceError: hadData is not defined 😆

PR #219, opened 2018-01-29
Backmerge: Fix Reducer Mutations

andre-meyer:
# Backmerge from Olympia
Fixes Reducer that caused mutations

### DIFF
--- a/src/reducers/entities.js
+++ b/src/reducers/entities.js
@@ -1,47 +1,39 @@
 import { handleActions } from 'redux-actions'
-import { keys, set, get } from 'lodash'
+import { keys, set, get, cloneDeep } from 'lodash'
 
 import {
   updateEntity,
   receiveEntities,
-  removeEntity,
 } from 'actions/entities'
 
 const reducer = handleActions({
   [updateEntity]: (state, action) => ({
-    ...state,
+    ...cloneDeep(state),
     [action.payload.entityType]: {
-      ...state[action.payload.entityType],
+      ...cloneDeep(state[action.payload.entityType]),
       [action.payload.data.id]: {
-        ...get(Object.assign({}, state), `[${action.payload.entityType}][${action.payload.data.id}]`, {}),
-        ...action.payload.data,
+        ...cloneDeep(get(state, `[${action.payload.entityType}][${action.payload.data.id}]`, {})),
+        ...cloneDeep(action.payload.data),
       },
     },
   }),
   [receiveEntities]: (state, action) => {
-    const result = { ...(Object.assign({}, state)) }
+    const result = cloneDeep(state)
     keys(action.payload.entities).forEach((entityType) => {
       // preserve old entities
-      const entitiesState = { ...(get(state, `${entityType}`, {})) }
+      const entitiesState = cloneDeep(get(state, `${entityType}`, {}))
       set(result, `${entityType}`, entitiesState)
 
       keys(action.payload.entities[entityType]).forEach((entityId) => {
         // preserve old entity (only update if not exists)
-        const entityState = get(Object.assign({}, state), `${entityType}.${entityId}`, {})
+        const entityState = cloneDeep(get(`${entityType}.${entityId}`, {}))
         set(result, `${entityType}.${entityId}`, {
           ...entityState,
-          ...Object.assign({}, action.payload.entities[entityType][entityId]),
+          ...cloneDeep(action.payload.entities[entityType][entityId]),
         })
       })
     })
     return result
-  },
-  [removeEntity]: (state, action) => {
-    const { ...rest } = state[action.payload.entityType]
-    return {
-      ...state,
-      [action.payload.entityType]: { ...rest },
-    }
   },
 }, {})
 
--- a/src/reducers/entities.js
+++ b/src/reducers/entities.js
@@ -13,24 +13,24 @@ const reducer = handleActions({
     [action.payload.entityType]: {
       ...state[action.payload.entityType],
       [action.payload.data.id]: {
-        ...get(state, `[${action.payload.entityType}][${action.payload.data.id}]`, {}),
+        ...get(Object.assign({}, state), `[${action.payload.entityType}][${action.payload.data.id}]`, {}),
         ...action.payload.data,
       },
     },
   }),
   [receiveEntities]: (state, action) => {
-    const result = { ...state }
+    const result = { ...(Object.assign({}, state)) }
     keys(action.payload.entities).forEach((entityType) => {
       // preserve old entities
-      const entitiesState = get(state, `${entityType}`, {})
+      const entitiesState = { ...(get(state, `${entityType}`, {})) }
       set(result, `${entityType}`, entitiesState)
 
       keys(action.payload.entities[entityType]).forEach((entityId) => {
         // preserve old entity (only update if not exists)
-        const entityState = get(state, `${entityType}.${entityId}`, {})
+        const entityState = get(Object.assign({}, state), `${entityType}.${entityId}`, {})
         set(result, `${entityType}.${entityId}`, {
           ...entityState,
-          ...action.payload.entities[entityType][entityId],
+          ...Object.assign({}, action.payload.entities[entityType][entityId]),
         })
       })
     })
@@ -40,7 +40,8 @@ const reducer = handleActions({
     const { ...rest } = state[action.payload.entityType]
     return {
       ...state,
-      [action.payload.entityType]: { ...rest } }
+      [action.payload.entityType]: { ...rest },
+    }
   },
 }, {})
 
--- a/src/reducers/entities.spec.js
+++ b/src/reducers/entities.spec.js
@@ -93,7 +93,7 @@ describe('Add a Market', () => {
 
   test('Update an existing market', () => {
     // GIVEN
-    const entityReducerWithMarket = {
+    const currentState = {
       eventDescriptions: { [TEST_EVENT_DESCRIPTION.ipfsHash]: TEST_EVENT_DESCRIPTION },
       oracles: { [TEST_ORACLE.address]: TEST_ORACLE },
       events: { [TEST_EVENT.address]: TEST_EVENT },
@@ -108,10 +108,10 @@ describe('Add a Market', () => {
     })
 
     // WHEN
-    const entityReducerWithUpdatedMarket = entityReducer(entityReducerWithMarket, updateMarketAction)
+    const updatedState = entityReducer(currentState, updateMarketAction)
 
     // THEN
-    const state = { entities: entityReducerWithUpdatedMarket }
+    const state = { entities: updatedState }
     const markets = getMarkets(state)
     expect(markets.length).toBe(1)
 
--- a/src/reducers/entities.spec.js
+++ b/src/reducers/entities.spec.js
@@ -58,7 +58,7 @@ const TEST_EVENT = {
   type: 'CATEGORICAL',
 }
 
-describe.only('Add a Market', () => {
+describe('Add a Market', () => {
   test('Add a market in an empty store', () => {
     // GIVEN
     const emptyEntityReducer = {}
--- a/src/reducers/entities.spec.js
+++ b/src/reducers/entities.spec.js
@@ -1,0 +1,101 @@
+import entityReducer from './entities'
+import { updateEntity, receiveEntities } from 'actions/entities'
+
+describe('entityReducer', () => {
+  describe('updateEntity', () => {
+    const currentEntityState = {
+      testEntity: {
+        'test-2': {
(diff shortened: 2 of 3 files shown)


src/reducers/entities.js
```diff
@@ -13,24 +13,24 @@ const reducer = handleActions({
...
       set(result, `${entityType}`, entitiesState)
 
       keys(action.payload.entities[entityType]).forEach((entityId) => {
         // preserve old entity (only update if not exists)
-        const entityState = get(state, `${entityType}.${entityId}`, {})
+        const entityState = get(Object.assign({}, state), `${entityType}.${entityId}`, {})
         set(result, `${entityType}.${entityId}`, {
           ...entityState,
-          ...action.payload.entities[entityType][entityId],
+          ...Object.assign({}, action.payload.entities[entityType][entityId]),
```
apanizo: Why are you using the spread operator before ```Object.assign```?

src/reducers/entities.js
```diff
@@ -13,24 +13,24 @@ const reducer = handleActions({
...
   [receiveEntities]: (state, action) => {
-    const result = { ...state }
+    const result = { ...(Object.assign({}, state)) }
     keys(action.payload.entities).forEach((entityType) => {
       // preserve old entities
-      const entitiesState = get(state, `${entityType}`, {})
+      const entitiesState = { ...(get(state, `${entityType}`, {})) }
       set(result, `${entityType}`, entitiesState)
 
       keys(action.payload.entities[entityType]).forEach((entityId) => {
```
apanizo: Why you just not build an object of entities and once you have finished iterating over them, you just include that new object to the state? In that way, you would save time and resources because if for any reason the state is too big and has other stuff not related with this "reducer" you are copying it in different places, right?

apanizo (commented):
I know this part will be refactored (introducing) immutable, so I have just asked a few question just in case I am missing something, I do not understand the reasons for those changes.

src/reducers/entities.js
```diff
@@ -13,24 +13,24 @@ const reducer = handleActions({
...
       },
     },
   }),
   [receiveEntities]: (state, action) => {
-    const result = { ...state }
+    const result = { ...(Object.assign({}, state)) }
     keys(action.payload.entities).forEach((entityType) => {
       // preserve old entities
-      const entitiesState = get(state, `${entityType}`, {})
+      const entitiesState = { ...(get(state, `${entityType}`, {})) }
```
apanizo: Why you need ```... ``` before the ```(get...)```?

andre-meyer: `get` returns a reference, this needs to be spread I think:
image

src/reducers/entities.js
mmvsha: Yeah, I've the same question. I think we can get rid of it here. Or maybe it will be fixed after the backmerge is done?

mmvsha:
```
var _extends = Object.assign || function (target) { for (var i = 1; i < arguments.length; i++) { var source = arguments[i]; for (var key in source) { if (Object.prototype.hasOwnProperty.call(source, key)) { target[key] = source[key]; } } } return target; };
```
This is a polyfill method from babel which is used to transpile spread operator. So you are doing `Object.assign({}, Object.assign({}, action.payload.entities[entityType][entityId]))`

andre-meyer: Can you explain to me how I can use `Object.assign` without spreading it? By putting everything into a variable first and then doing Object.assign on it?

apanizo: https://redux.js.org/docs/recipes/UsingObjectSpreadOperator.html

If it is not clear enough, let me know and I will provide you more concrete examples.

src/reducers/entities.js
apanizo: Normally I am used to see, new data coming actions, and update something in the store at once using ```...``` or ```spread operator``` in this case it seems you need old data from the state as well, I suggest follow the same pattern. Fetch/query data (using lodash), and then at the end build the new state using: ```...oldState, dataFromAction, ...dataFetchedFromState```

Do you understand what I mean?

andre-meyer:
I refactored this (briefly) implementing tests and using `cloneDeep` from lodash to ensure there's no more mutations

mmv08 (approved):

src/reducers/entities.spec.js
```diff
@@ -0,0 +1,101 @@
+import entityReducer from './entities'
+import { updateEntity, receiveEntities } from 'actions/entities'
+
+describe('entityReducer', () => {
```
apanizo: Ideally, tests should make easier understand what the code does, where the code is used, and what are corner cases of those use cases.

For doing that, we need to make them more readable. Somethin like should work:
```
// entities.spec.js  <- Good name. it specifies what we are testing
describe('Reducer Entities: Adding users')  <-- Should describe a feature of the domain we are testing
    test("Adding users when the store is empty")   <-- Should specify all different cases
        ...
    test("Adding users different users")
        ...
    test("Modifying existing users")  <-- This test is an example of the typical test added later in time
                                                                  because we have had a bug, like mutations
        //GIVEN
             const storeWithUsers  //names are just an example
             const dataFromEndpointWithSameUsers  //names are just an example
        //WHEN
             //adding to the store              
        //THEN
            //We do not have mutations 
```
If you follow this pattern, you will have better tests, more organised and new developers will understand everything much better and faster.

andre-meyer: Thanks for the feedback, can you maybe give me a concrete example of how this should look for this reducer?

I rewrote it like this, wdyt?
```jsx
describe('Reducer Entities: Adding and updating Entities', () => {
  describe('Action: updateEntity', () => {
    // GIVEN
    const stateWithTestEntity = {
      testEntity: {
        'test-2': {
          id: 'test-2',
          deeplyNestedData: {
            deeplyNestedKey: 'moreNestedData',
          },
        },
      },
    }

    // WHEN
    const updateAction = updateEntity({ entityType: 'testEntity', data: { id: 'test-1', testObject: { deeplyNested: 'aNestedVaue' } } })
    const stateWithUpdatedEntity = entityReducer(stateWithTestEntity, updateAction)

    // THEN
    test('Updating an Entity', () => {
      // Ensure Action Payload matches Entity Payload
      expect(stateWithTestEntity.testEntity['test-2'].deeplyNestedData).toMatchObject(stateWithUpdatedEntity.testEntity['test-2'].deeplyNestedData)
      expect(updateAction.payload.data.testObject).toMatchObject(stateWithUpdatedEntity.testEntity['test-1'].testObject)
      
      // Ensure no mutations! (=== Compare)
      expect(stateWithTestEntity.testEntity['test-2'].deeplyNestedData).not.toBe(stateWithUpdatedEntity.testEntity['test-2'].deeplyNestedData)
      expect(updateAction.payload.data.testObject).not.toBe(stateWithUpdatedEntity.testEntity['test-1'].testObject)

      // Ensure new Entity was added
      const stateWithNewEntity = {
        'test-2': {
          id: 'test-2',
          deeplyNestedData: {
            deeplyNestedKey: 'moreNestedData',
          },
        },
        'test-1': {
          id: 'test-1',
          testObject: { deeplyNested: 'aNestedVaue' },
        },
      }
      expect(stateWithUpdatedEntity.testEntity).toMatchObject(stateWithNewEntity)
    })
  })
```

apanizo: Ok, let me ask first a couple of questions:
+  Why do you think it is a good idea to have chained ```describe```? 
+  Which entities are you testing? Users? Trades? Markets?

andre-meyer: - I'm testing a reducer and inside this reducer, the two specified actions it can receive
- I'm testing the entity reducer, so I'm using an entity called "TestEntity"

andre-meyer: @apanizo 
Okay, so does this make more sense then?
```javascript
describe('Reducer Entities: Adding and updating Entities', () => {
  test('Updating an Entity', () => {
    // GIVEN
    const stateWithTestEntity = {
      testEntity: {
        'test-2': {
          id: 'test-2',
          deeplyNestedData: {
            deeplyNestedKey: 'moreNestedData',
          },
        },
      },
    }

    // WHEN
    const updateAction = updateEntity({ entityType: 'testEntity', data: { id: 'test-1', testObject: { deeplyNested: 'aNestedVaue' } } })
    const stateWithUpdatedEntity = entityReducer(stateWithTestEntity, updateAction)

    // THEN
    // Ensure Action Payload matches Entity Payload
    expect(stateWithTestEntity.testEntity['test-2'].deeplyNestedData).toMatchObject(stateWithUpdatedEntity.testEntity['test-2'].deeplyNestedData)
    expect(updateAction.payload.data.testObject).toMatchObject(stateWithUpdatedEntity.testEntity['test-1'].testObject)
    
    // Ensure no mutations! (=== Compare)
    expect(stateWithTestEntity.testEntity['test-2'].deeplyNestedData).not.toBe(stateWithUpdatedEntity.testEntity['test-2'].deeplyNestedData)
    expect(updateAction.payload.data.testObject).not.toBe(stateWithUpdatedEntity.testEntity['test-1'].testObject)

    // Ensure new Entity was added
    const stateWithNewEntity = {
      'test-2': {
        id: 'test-2',
        deeplyNestedData: {
          deeplyNestedKey: 'moreNestedData',
        },
      },
      'test-1': {
        id: 'test-1',
        testObject: { deeplyNested: 'aNestedVaue' },
      },
    }
    expect(stateWithUpdatedEntity.testEntity).toMatchObject(stateWithNewEntity)
  })
```

apanizo: Testing should be done based on features. We can say a feature is something that the user can do in the application.

Orienting tests in that way will help to understand what the code should do and what we expect from that code.  So Testing a reducer is not something a user can do.

So the question is... what the user can do for triggering the usage of that reducer? Maybe load markets and therefore "Add Markets" to the store? Maybe delete a market? 

With that information, I will compose the "describe" functions. For example:
```
describe('Add Markets to store')
describe('Remove markets to store')
```
Once I have the features _described_ the next step is to write the different tests cases. For example:
Which different options can we have when we add markets?
  * Adding markets when the store is empty ```test('Adding markets when store is empty`)```
  * Adding different markets that the ones who are inside the store
  * Finally modify existing markets

The last point is (or the last test case) is the typical test you do not add at the beginning, is the typical test you add when you have had a bug (like mutations).

The last step is how to organize the test. For reading purposes I suggest to use:
```
GIVEN
WHEN
THEN
```
approach.

For example:
```
describe('Add Markets') { 
    test('Add Markets when store is empty') {
        //GIVEN
        const emptyStore = ...;
        const payload = request.payload;

        //WHEN... adding markets to store
        const store = {...emptyStore, payload.markets }

       //THEN... we have the markets in the store
        const markets = getMarketSelector()
       testEqual(market[0], payload.market[0]);
       testEqual(market[1], payload.market[1]);
       ....
    }
    ....
}
describe(' ...') {
    ....
}
.....
```
So I do not think having chained is good in terms of understanding correctly what the app does.

src/reducers/entities.spec.js
```diff
@@ -0,0 +1,131 @@
...
+    expect(market.event).toMatchObject(TEST_EVENT)
+
+    expect(market.oracle).not.toBe(TEST_ORACLE) // no mutation check
+    expect(market.eventDescription).not.toBe(TEST_EVENT_DESCRIPTION) // no mutation check
+    expect(market.event).not.toBe(TEST_EVENT) // no mutation check
+  })
+
+  test('Update an existing market', () => {
+    // GIVEN
+    const entityReducerWithMarket = {
```
apanizo: I would say: ```currentState```

src/reducers/entities.spec.js
```diff
@@ -0,0 +1,131 @@
...
+    const updateMarketAction = updateEntity({
+      entityType: 'oracles',
+      data: {
+        id: TEST_ORACLE.address,
+        creator: '0xf00',
+      },
+    })
+
+    // WHEN
+    const entityReducerWithUpdatedMarket = entityReducer(entityReducerWithMarket, updateMarketAction)
```
apanizo: ... and ```updatedState```

src/reducers/entities.spec.js
```diff
@@ -0,0 +1,131 @@
...
+
+    // THEN
+    const state = { entities: entityReducerWithUpdatedMarket }
+    const markets = getMarkets(state)
```
apanizo: When refactoring we need to investigate how hard would inject a "real" redux state.

apanizo (approved):
Really good job. Now if someone needs to understand what the entities reducer does, if he goes to the test he will understand it is used in markets and will understand we have had mutations problems (which they are already fixed).

Probably I would rename a couple of variables, but it looks good to me. 

Congrats. Good job.

PR #222, opened 2018-02-05
PM-436 Refactor: Update dependencies 

mmvsha:


src/components/Header/index.js
```diff
@@ -27,38 +27,38 @@ class Header extends Component {
...
-            </Link>
+            </NavLink>
           </div>
           <div className="headerContainer__group headerContainer__group--left headerContainer__group--version">
             {version}
           </div>
           <div className="headerContainer__group headerContainer__group--left navLinks">
             {hasWallet && (
-              <Link
+              <NavLink
```
apanizo: Why the ```NavLink```?

src/index.js
```diff
@@ -4,17 +4,16 @@ import Raven from 'raven-js'
...
-import { browserHistory } from 'react-router'
-import { syncHistoryWithStore } from 'react-router-redux'
 import { AppContainer } from 'react-hot-loader'
 import 'less/style.less'
-import AppRouter from 'router'
+import AppRouter from 'routes'
 import initGoogleAnalytics from 'utils/analytics/init'
 import BackdropProvider from 'containers/BackdropProvider'
-import store from 'store'
+import store, { history } from 'store'
```
apanizo: is this the same as:

``` 
import { createBrowserHistory } from 'history';

const history = createBrowserHistory();
```
? 

I do not exactly remember but I think it is better to go with this option. Can you have a look?

src/components/Header/index.js
mmvsha: @apanizo We're using `activeClassName` prop and it's not on `Link` anymore.

><NavLink> works the same as <Link> but gives you some extra styling abilities depending >on if the <NavLink> matches the browser's URL.
https://css-tricks.com/react-router-4/

src/index.js
mmvsha: @apanizo It's created inside store.js because we need to pass it into the middleware. I thought that we have to use the same history object everywhere, right? This is why I did it

PR #227, opened 2018-02-07
PM-439 Bug: Do not run watcher if the provider hasn't tried to init

mmvsha:
So the reason for wrong network modal showing up was a watcher for changes. It had run before a provider tried to init and those lines of code made the provider available:
```
    if (!this.walletEnabled) {
      this.walletEnabled = true
      providerUpdate.available = true
    }
```

I added connectionTried variable and now watcher checks if provider has actually tried to connect.


apanizo:
I do not like in general condition races and its hacks. It should be a real problem if we load providers and then, once they are loaded, we execute the watcher?

mmvsha:
@apanizo Ok, will redo it

apanizo:
Before spending too much time, think about it first, if you have any problem let me know.

src/integrations/metamask/index.js
```diff
@@ -50,6 +50,10 @@ class Metamask extends InjectedWeb3 {
       this.walletEnabled = false
     }
 
+    if (this.watcher) {
```
apanizo: Code LGTM but I suggest to refactor the the ```this.walletEnabled``` stuff. This file is 80 LOC and it would not be a problem if you just create a function for that. Something like:
```
this.walletEnabled = initWallet()
....
const initWallet = () => {
  try {
    if (!...) {
        return false
    }
    // init provider
} catch (err) {
    return false
}
```

mmvsha:
@apanizo 👌

PR #230, opened 2018-02-09
Feature PM-411 creation action reducers market list route

apanizo:
**Due Diligence**
- [ ] Breaking change
- [x] Tests //JEST, Storybook, Chromatic components
- [ ] Documentation

**Description**
This PR creates the market list route (folders). Also, it includes the reducer and necessary actions for fetching, manipulate and store data (simplified) in the store. 

It includes tests based on features, using the "Builder" pattern. Those tests simulate a "real" redux store.

**DevNotes**
Additionally, it includes an update to Babel 7 and a simplified configuration for running tests on Travis.

The Builder pattern allows us to create markets in this way:
```
cons aEthereumMarket = aMarket()
    .withTitle('What will the number of Ethereum transactions be on January 3rd, 2018?')
    .withDate('2018-01-03T12:00:00')
    .withVolume('857142857142872776')
    .withType(MARKET_SCALAR)
    .withOutcomes()
    .withBounds('500000', '1250000', 'Txs')
    .get()
```
Which simplifies everything for following the GIVEN-WHEN-THEN approach.

**PM Notes**
This PR ticket is a subtask of PM-409

### DIFF
--- a/src/routes/marketlist/store/models/bounds.js
+++ b/src/routes/marketlist/store/models/bounds.js
@@ -1,9 +1,9 @@
 import { Record } from 'immutable'
 
-const Bound = Record({
+const BoundsRecord = Record({
   lower: undefined,
   upper: undefined,
   unit: undefined,
 }, 'Bound')
 
-export default Bound
+export default BoundsRecord
--- a/src/routes/marketlist/store/models/outcome.js
+++ b/src/routes/marketlist/store/models/outcome.js
@@ -1,8 +1,8 @@
 import { Record } from 'immutable'
 
-const Outcome = Record({
+const OutcomeRecord = Record({
   name: undefined, // string
   marginalPrice: undefined, // decimal
 }, 'Outcome')
 
-export default Outcome
+export default OutcomeRecord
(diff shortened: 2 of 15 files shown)


src/routes/marketlist/store/models/bounds.js
```diff
@@ -0,0 +1,9 @@
+import { Record } from 'immutable'
```
apanizo: @mmvsha  @andre-meyer 
Maybe should I introduce this record inside ```store/models/market.js```?

src/routes/marketlist/store/models/outcome.js
```diff
@@ -0,0 +1,8 @@
+import { Record } from 'immutable'
```
apanizo: @mmvsha @andre-meyer
Maybe should I introduce this record inside store/models/market.js?

apanizo:
@mmvsha  @andre-meyer As you can see ```src/routes/marketlist/store/reducers/market.builder.js``` is a file which abstract some things done in the tests (like the MarketBuilder or objects simulating real data).

Do you have any suggestion to split this file? Keep in mind this file is just for testing and should be skipped when we create the prod build.

src/routes/marketlist/store/models/bounds.js
mmvsha: I'd say yes, the file is not big and it's not a problem to have multiple entities in the same file
@apanizo

apanizo: 👍

PR #231, opened 2018-02-12
Refactor PM-440: Integrations, blockchain

mmvsha:
This PR includes:
- Integrations and blockchain stuff are now split
- Reducers, selectors are now using immutable.js
- Provider integrations related reducers, actions, etc are now in `integrations/store` folder
- Import paths are not absolute and not relative everywhere I was working

Please review it after PM-439, I needed some changes from it so it includes changes from PM-439


src/containers/App/index.js
```diff
@@ -13,7 +13,8 @@ import { providerPropType } from 'utils/shapes'
 
 import HeaderContainer from 'containers/HeaderContainer'
 import TransactionFloaterContainer from 'containers/TransactionFloaterContainer'
-import { getSelectedProvider, isConnectedToCorrectNetwork } from 'selectors/blockchain'
+import { triedToConnect } from 'selectors/blockchain'
```
andre-meyer: @mmvsha Is this still a general folder for blockchain related selectors?

PR #232, opened 2018-02-12
BUG PM-442: Fix redirection problems

mmvsha:
This PR fixes:
- Not redirecting from `markets/*invalid id*` paths
- App crash if redirecting multiple times to `/` and then to `markets/list`

### DIFF
--- a/src/routes/index.js
+++ b/src/routes/index.js
@@ -1,5 +1,6 @@
 import React from 'react'
 import { Switch, Route, Redirect } from 'react-router-dom'
+import web3 from 'web3'
 
 import MarketListPage from 'containers/MarketListPage'
 import MarketDetailPage from 'containers/MarketDetailPage'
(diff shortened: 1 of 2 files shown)


src/routes/index.js
```diff
@@ -12,7 +13,16 @@ const AppRouter = () => (
...
-    <Route exact path="/markets/:id/:view?/:shareId?" component={MarketDetailPage} />
+    <Route
+      exact
+      path="/markets/:id/:view?/:shareId?"
+      render={(props) => {
+        if (web3.utils.isAddress(props.match.params.id)) {
+          return <MarketDetailPage {...props} />
+        }
+        return <Redirect to="/markets/list" />
+      }}
```
andre-meyer: Can you move this function out of the route definiton into a seperate variable?

mmvsha: @andre-meyer Sorry, don't get it. What do you mean?

andre-meyer:
```const marketDetailRender = (props) => {
        if (web3.utils.isAddress(props.match.params.id)) {
          return <MarketDetailPage {...props} />
        }
        return <Redirect to="/markets/list" />
      }}```

PR #239, opened 2018-02-15
Refactor Formcomponents: Slider

andre-meyer:
# Belongs to PM-412

image


mmvsha:
Hmm, I think the text inside input should have a little bit more margin-left. What do you think?

mmv08 (approved):

PR #240, opened 2018-02-15
Refactor Formcomponents: Textarea

andre-meyer:
# Belongs to PM-412

image

### DIFF
--- a/src/components/Form/TextArea.js
+++ b/src/components/Form/TextArea.js
@@ -16,18 +16,21 @@ const TextArea = ({
   label,
   className,
   meta: { touched, error },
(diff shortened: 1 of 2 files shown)


mmvsha:
I think some time ago we agreed that we should have an option to format text. I think this is perfect time to do it. For example, in Olympia we had links in market description and they were not clickable. What do you think?

mmvsha:
Oh wait, why do we need this component if market creation was removed (I think it was used only here)

andre-meyer:
Good point, only used in Market Creation. I'll close this PR :)

PR #245, opened 2018-02-20
Refactor PM-407: Outcome components

mmvsha:
This PR:
- Moves all outcome components to `components/Outcome`
- Adds `TrendingOutcomeScalar` and `TrendingOutcomeCategorical`
- Uses CSS modules in all outcome components now
- Adds defaultProps
- Prepares `webpack.config.dev.js` for CSS modules


src/components/Outcome/OutcomeCategorical/TrendingOutcomeCategorical/index.js
```diff
@@ -0,0 +1,35 @@
...
+import style from '../outcomeCategorical.scss'
+
+const cx = cn.bind(style)
+
+const TrendingOutcomeCategorical = ({
+  entryStyle, outcome, percentage, resolutionDate,
+}) => (
+  <div className="outcomes outcomes--categorical">
+    <div className={cx('outcomeWrapper')}>
+      <div className="entry__color" style={entryStyle} />
```
andre-meyer: Why not use `cx('entry')` here?

src/components/Outcome/OutcomeCategorical/TrendingOutcomeCategorical/index.js
```diff
@@ -0,0 +1,35 @@
...
+import PropTypes from 'prop-types'
+import cn from 'classnames'
+import style from '../outcomeCategorical.scss'
+
+const cx = cn.bind(style)
+
+const TrendingOutcomeCategorical = ({
+  entryStyle, outcome, percentage, resolutionDate,
+}) => (
+  <div className="outcomes outcomes--categorical">
```
andre-meyer: `cx('outcomes', 'categorical')`

if we're going for css-modules, we need to use the bound styles everywhere

webpack.dev.config.js
```diff
@@ -60,6 +60,8 @@ module.exports = {
             loader: 'css-loader',
             options: {
               sourceMap: true,
+              importLoaders: 2,
```
andre-meyer: What does this do?

webpack.dev.config.js
```diff
@@ -60,6 +60,8 @@ module.exports = {
...
+              localIdentName: '[name]--[local]--[hash:base64:8]',
```
andre-meyer: Doesn't this break all previously declared global css?

webpack.dev.config.js
mmvsha: @andre-meyer https://github.com/webpack-contrib/css-loader/issues/228

src/components/Outcome/OutcomeCategorical/TrendingOutcomeCategorical/index.js
mmvsha: @andre-meyer Will fix that, this is related to the question I asked you today about adding class names for elements with no style imported from .scss

webpack.dev.config.js
mmvsha: @andre-meyer Right now it doesn't, it will break after we set `modules: true`, this is just some kind of preparation for names 😃

src/components/Outcome/OutcomeCategorical/TrendingOutcomeCategorical/index.js
mmvsha: 👌

PR #246, opened 2018-02-20
Feature PM-466: Fix the tooltip for the wrong chain

mmvsha:
**BEFORE:** 
screen shot 2018-02-20 at 17 10 57

**AFTER:**
screen shot 2018-02-20 at 17 10 34

**ALSO:**
- Use css modules
- Add default props


src/containers/InteractionButton/index.js
```diff
@@ -108,7 +110,11 @@ class InteractionButton extends Component {
 
     if (networkError) {
       return (
-        <Tooltip overlay="You are connected to the wrong chain. You can only interact with Gnosis when you're on the same chain as us.">
+        <Tooltip
+          overlay={`You are connected to the wrong chain. You can only interact using ${
+            ETHEREUM_NETWORK_IDS[targetNetworkId]
```
andre-meyer: Please display the network in normal casing using `upperFirst(networkName.toLowerCase()`: https://lodash.com/docs/4.17.4#upperFirst

Maybe move to a separate variable for the overlay text

PR #249, opened 2018-02-21
Feature PM-480 container market list route

apanizo:
This PR contains the logic for using the new market list container. IT includes the redux store market list reducer and also the update of the RR4 route.

NOTE: I have left the old container just for not breaking the UI, next step is implement the MarketList component (```src/routes/marketlist/components/Layout```).

See:
screenshot 2018-02-21 15 11 18

screenshot 2018-02-21 15 11 35

screenshot 2018-02-21 15 56 25


src/routes/marketlist/components/Layout.js
```diff
@@ -0,0 +1,7 @@
+import React from 'react'
+
+const Layout = ({openMarkets, newMarkets, endingSoonMarkets}) => (
```
apanizo: @andre-meyer @mmvsha That is the file to update in the next step.

src/routes/index.js
```diff
@@ -3,6 +3,7 @@ import { Switch, Route, Redirect } from 'react-router-dom'
 import web3 from 'web3'
 
 import MarketListPage from 'containers/MarketListPage'
+// import MarketListPage from 'routes/marketlist/container/MarketList'
```
apanizo: Once the Layout is implemented properly in the next step we should use this reference.

PR #250, opened 2018-02-22
Task: Fix CSS-Modules Chaos - Global by default

andre-meyer:
To solve our current problems with only having partial css-modules support I added a rule to ensure there are no duplications from global classes when working with css-modules.

In order to use css-modules with the current Webpack Configuration, simply append .mod before the extension, so **for example `Component.mod.scss` this will enable local variable names and avoid problems with interfering global css**. In the future we will rename all .mod files back to .scss and only use css-modules.

This functionality can also be used to test how components will look like if only css-modules was allowed, because when classnames are bound (using `classNames.bind(css)`), it won't allow global css anymore.

If you can think of a better solution, let me know!


mmvsha:
I have a question, will it prevent us from using bootstrap classes?

andre-meyer:
No, you can simply use them like before doing this:
`className="container"`
or if you're using it together with css-module classes:
`className={cx('cssModuleClass', 'container')}` (this works because as long as `container` is not defined in the css-modules stylesheet, it will not match a *local ident name* and use the global bootstrap css. :)

PR #258, opened 2018-03-09
Task: Fix Form Issues

andre-meyer:
Thanks to Mikhail for pointing out these issues:

- TextInput now has a `decoration` prop where you can pass `underlined` and it will have that dotted underline that we used everywhere
- Updated the OutcomeSelection to have a `hideBars` prop to disable the actual outcome probability bars on scalar markets

- I didn’t change the slider, I think it looks better this way, the previous version was more inconsistent across the screens, the `light` version of the scalar is supposed to look like this imo, otherwise let's change this in the design overhaul

### DIFF
--- a/src/components/Form/InputError.mod.scss
+++ b/src/components/Form/InputError.mod.scss
@@ -1,6 +1,7 @@
 @import "../../scss/vars.scss";
 
 .inputError {
+  position: absolute;
   font-weight: 800;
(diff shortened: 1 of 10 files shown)


mmv08 (approved):

mmvsha:
buuuut why did the tests fail?

PR #269, opened 2018-03-22
Task PM-554: Configure registration method

mmvsha:


src/api/account.js
```diff
@@ -23,26 +22,6 @@ export const getCurrentBalance = async (account) => {
...
-
-export const getOlympiaTokensByAccount = async (account) => {
-  const gnosis = await api.getGnosisConnection()
-  const balance = await gnosis.olympiaToken.balanceOf(account)
-  return new Decimal(balance.toFixed(0)).toString()
-}
-
 export const getMainnetAddressForRinkebyAccount = async (account) => {
   const gnosis = await api.getGnosisConnection()
   const address = await gnosis.olympiaAddressRegistry.mainnetAddressFor(hexWithPrefix(account))
```
andre-meyer: This should be configurable too. Wdyt?

src/config.json
```diff
@@ -42,6 +42,13 @@
         "markdown": true
       }
     },
-    "faviconPath": "assets/img/gnosis_logo_icon.svg"
+    "faviconPath": "assets/img/gnosis_logo_icon.svg",
+    "provider": "UPORT",
```
andre-meyer: Is this a `defaultProvider` maybe?

src/integrations/uport/index.js
```diff
@@ -0,0 +1,116 @@
+import { getTokenBalance } from 'api'
+import { setConnectionStatus } from 'actions/blockchain'
+import { WALLET_PROVIDER } from 'integrations/constants'
+import InjectedWeb3 from 'integrations/injectedWeb3'
+import { fetchTournamentUserData } from 'routes/Scoreboard/store/actions'
+import { weiToEth, hexWithoutPrefix } from 'utils/helpers'
+import { getTokenAddress } from 'utils/configuration'
+import initUportConnector, { connect, connectorLogOut, isUserConnected } from './connector'
+
+export const notificationsEnabled = false
```
andre-meyer: Use config here? Maybe it would make sense to turn the `provider` key into an object, supplying a `defaultProvider` and `options` that can be used in the actual integrations (and would be specific to them)

src/api/account.js
mmvsha: Will do it later in user registration and reward claiming ticket @andre-meyer

src/config.json
mmvsha: @andre-meyer For me "defaultProvider" implies that there will be more providers than just one. Here we'll use the only provider specified. hope you get it

PR #274, opened 2018-04-02
Task PM-517: User registration

mmvsha:


src/components/ModalContent/SwitchNetwork/index.js
```diff
@@ -0,0 +1,31 @@
...
+const logoStyle = {
+  width: 100,
+  height: 100,
+}
+
+const SwitchNetwork = ({ closeModal, targetNetwork }) => (
+  <div className={cx('switchNetwork')}>
+    <button className={cx('closeButton')} onClick={() => closeModal()} />
+    <img src={MetamaskIcon} alt="logo" style={logoStyle} />
+    <h3 className={cx('switchNetworkHeading')}>Switch to the {targetNetwork} Network</h3>
```
andre-meyer: a classname of `heading` would make this more readable, and inside the scss file you can simply nest a `.heading` rule inside the `.switchNetwork` style and it won't interfere with global css.

src/components/ModalContent/SwitchNetwork/index.js
```diff
@@ -0,0 +1,31 @@
...
+    <p className={cx('switchNetworkText')}>
```
andre-meyer: This can also be `text`

src/components/ModalContent/UnlockMetamask/index.js
```diff
@@ -0,0 +1,27 @@
...
+const logoStyle = {
+  width: 100,
+  height: 100,
+}
+
+const UnlockMetamask = ({ closeModal }) => (
+  <div className={cx('unlockMetamask')}>
+    <button className={cx('closeButton')} onClick={closeModal} />
+    <img src={MetamaskIcon} alt="logo" style={logoStyle} />
+    <h3 className={cx('unlockHeading')}>Unlock your MetaMask wallet</h3>
```
andre-meyer: `.heading`?

src/components/ModalContent/UnlockMetamask/index.js
```diff
@@ -0,0 +1,27 @@
...
+    <p className={cx('unlockText')}>Please unlock your MetaMask wallet to start using Gnosis Olympia.</p>
```
andre-meyer: `.text`?

src/routes/MarketDetails/store/actions/buyShares.js
```diff
@@ -60,7 +60,8 @@ const buyMarketShares = (market, outcomeIndex, outcomeTokenCount, cost) => async
   }
   if (approvalResetAmount) transactions.unshift(SETTING_ALLOWANCE)
 
-  const payload = await api.requestMarket(market.address)
+  console.log(fetchMarket)
```
andre-meyer: `console.log`

src/components/Header/index.js
```diff
@@ -17,9 +19,37 @@ import css from './Header.scss'
...
+    const { isConnectedToCorrectNetwork, lockedMetamask } = this.props
+    if (!hasMetamask() && !shouldUseUport) {
+      this.props.openInstallMetamaskModal()
+    } else if (shouldUseMetamask()) {
+      if (lockedMetamask) {
+        this.props.openUnlockMetamaskModal()
+      } else if (!isConnectedToCorrectNetwork) {
+        this.props.openSwitchNetworkModal()
+      } else {
+        this.props.openRegisterWalletModal()
```
andre-meyer: I think what would be nice here for all the modal functions, would be a function called `openModal` where you simply pass the name of the modal (and additional parameters to display it) instead of having very verbose action functions for each modal. What do you think?

src/components/Header/index.js
```diff
@@ -38,8 +68,14 @@ class Header extends Component {
...
     } = this.props
 
+    let walletConnected = hasWallet
+    if (isTournament && shouldUseMetamask()) {
+      walletConnected = hasWallet && mainnetAddress
+    }
+
     const logoVars = {}
     if (isTournament) {
       logoVars['--logoAnnotation'] = "'Powered by Gnosis'"
```
andre-meyer: I know this is unrelated to the PR but can you explain this please? Why the double hyphen?

src/integrations/uport/index.js
```diff
@@ -70,7 +70,7 @@ class Uport extends InjectedWeb3 {
   }
 
   getNetworkId() {
-    return hexWithoutPrefix(this.web3.currentProvider.networkId)
+    return +hexWithoutPrefix(this.web3.currentProvider.networkId)
```
andre-meyer: Do you need it as a number? I think this can be done more nicely. Why do you need it to be a number and where?

src/components/Header/index.js
mmvsha: CSS vars start with double hyphen. I did it with CSS variables because JS cannot do `import X from variable` or `const X = require(variable)` and we have to use different images. Maybe this could be done in other way - import multiple IMGs and then conditionally use one of them. What do you think?

src/integrations/uport/index.js
mmvsha: `IsConnectedToCorrectNetwork` selector compares it against a number @andre-meyer

src/components/Header/index.js
mmvsha: @andre-meyer

src/components/Form/TextInputAdornment.js
```diff
@@ -6,14 +6,14 @@ import styles from './TextInputAdornment.mod.scss'
...
-const TextInputAdornment = ({ children, position }) => {
+const TextInputAdornment = ({ children, position, style }) => {
   const textInputAdornmentClasses = cx('adornment', {
     end: position === 'end',
     start: position === 'start',
   })
 
   return (
-    <p className={textInputAdornmentClasses}>
+    <p className={textInputAdornmentClasses} style={style}>
```
andre-meyer: Can we avoid this here? Can you maybe instead use a prop to set a certain type of style?

andre-meyer: Actually this is not used anywhere else yet, so feel free to set the `padding: 0` from the MarketDetail component that you used in the TextInputAdornments styles and remove the `style` prop

PR #275, opened 2018-04-05
Create LICENSE

mmvsha:


andre-meyer:
I think it needs to be a .md file so github correctly parses it including headings and whatnot: https://github.com/reactjs/redux/blob/master/LICENSE.md

mmvsha:
@andre-meyer I created this via github. i don't know why it did it this way. Let's postpone this PR for a little

PR #280, opened 2018-04-10
Add Read-only gnosis.js instance

mmv08:


andre-meyer:
Fixed with https://github.com/gnosis/gnosis-management/pull/282

PR #281, opened 2018-04-11
Dashboard: Configurable Token Icon and various style fixes

andre-meyer:
Added config entry for 'tokenIcon'


.nvmrc
```diff
@@ -1 +1 @@
-v8.1.0
+9.11.1
```
mmvsha: shouldn't it be ignored?

mmv08 (approved):

.nvmrc
andre-meyer: No, `.nvmrc` defines which version should be used when running the project. Of course it can be ignored though. I couldn't compile with <9.0 with yarn so I updated to 9.11.1

PR #282, opened 2018-04-11
Task: Enable readOnly Remote Provider

andre-meyer:
Allows reading from blockchain without allowing interaction through contracts.

### DIFF
--- a/src/actions/blockchain.js
+++ b/src/actions/blockchain.js
@@ -1,6 +1,7 @@
 import {
   initGnosisConnection,
-  getCurrentBalance,
+  // getCurrentBalance,
+  getIsListening,
   getCurrentAccount,
   getGasPrice,
   getTokenSymbol,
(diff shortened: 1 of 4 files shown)


src/actions/blockchain.js
```diff
@@ -1,6 +1,7 @@
 import {
   initGnosisConnection,
-  getCurrentBalance,
+  // getCurrentBalance,
```
mmvsha: why did you comment it?

src/integrations/remote/index.js
```diff
@@ -56,15 +54,18 @@ class Remote extends InjectedWeb3 {
...
-      network: this.network,
-      account: this.account,
-      balance: this.balance,
-    })
-      .then(() => opts.initGnosis())
-      .catch(() => opts.initGnosis())
+    try {
+      await this.runProviderUpdate(this, {
+        available: this.walletEnabled && this.account != null,
+        readOnly: this.walletEnabled,
```
mmvsha: shouldn't `readOnly` prop take into account if there is an account available?

PR #307, opened 2018-04-23
Refactor components to use CSS Modules

mmvsha:


src/components/Icon/index.js
```diff
@@ -7,14 +7,14 @@ import css from './Icon.mod.scss'
 const cx = classNames.bind(css)
 
 const Icon = ({
-  src, type, size, float, style, ...props
+  src, type, size, float, style, className, ...props
```
andre-meyer: Do we need className here? I'd prefer to style this component only via components and not via passed classes

PR #309, opened 2018-04-24
Task: Fix CollateralToken issues on non-tournament mode

andre-meyer:
CollateralToken is now an object that can be requested using the selector in /selectors/blockchain. It returns the collateralTokens symbol, amount in ETH, icon and address

### DIFF
--- a/src/selectors/blockchain.js
+++ b/src/selectors/blockchain.js
@@ -1,4 +1,9 @@
 import Decimal from 'decimal.js'
+import { getCollateralToken } from 'utils/features'
+import { getCurrentBalance } from 'integrations/store/selectors'
+import { weiToEth } from 'utils/helpers'
+
+const ETH_TOKEN_ICON = 'assets/img/icons/icon_etherTokens.svg'
 
 /**
  * Returns if gnosis.js is initialized or not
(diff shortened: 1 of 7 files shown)


src/selectors/blockchain.js
```diff
@@ -23,3 +28,27 @@ export const getTokenAmount = (state, tokenAddress) => {
 }
 
 export const getTokenSymbol = (state, tokenAddress) => state.blockchain.getIn(['tokenSymbols', tokenAddress])
+
+/**
+ * @param {*} state - redux state
+ * @returns {object} collateralToken - with Symbol, Amount in ETH and Address
+ */
+export const getCollateralTokenAmount = (state) => {
```
mmvsha: Isn't the name misleading? It returns not only amount, but symbol, address, icon too :) Maybe `getCollateralTokenInfo` would be a better name

src/selectors/blockchain.js
```diff
@@ -23,3 +28,27 @@ export const getTokenAmount = (state, tokenAddress) => {
...
+  const collateralToken = getCollateralToken()
+
+  if (!collateralToken) {
+    return {
+      symbol: 'ETH',
+      amount: weiToEth(getCurrentBalance(state)),
+      address: undefined,
```
mmvsha: what about returning an address of deployed etherToken from gnosis.js?
```
address: gnosis.etherToken.address
```

andre-meyer: That'd be wrapped ether, the header should display just ether for now

mmv08 (approved):

PR #324, opened 2018-05-07
Task PM-602: Add ToS, PP and RDP to set mainnet address modal

mmvsha:
### Description
* Update `webpack` and `webpack-cli`
* Add ToS, PP and RDP checkboxes to set mainnet address modal
* Fix footer
- Tried different configurations (specified/unspecified ToS, PP url) and made sure it works okay


webpack.prod.config.js
```diff
@@ -139,6 +140,7 @@ module.exports = (env = {}) => {
         },
       }),
       new webpack.ContextReplacementPlugin(/moment[/\\]locale$/, /en/),
+      new CopyWebpackPlugin(['assets/TermsOfService.html', 'assets/PrivacyPolicy.html', 'assets/RiskDisclaimerPolicy.html']),
```
andre-meyer: @mmvsha I think this is something we could easily forget. Can you move text content into a folder from root: `/content` and copy those into the build? Wdyt?

PR #330, opened 2018-05-10
Task PM-612: Filter markets by collateral token address from config

mmvsha:
### Description
* Add filter to `processMarketResponse` if a collateral token specified in config

* Tested multiple collateral token in config and made sure everything is working

### Background Information
* I wasn't really sure if the right place to do it is selectors (previously filters were done there) or response processing, I chose response processing because I feel like we don't want to save unfiltered markets at all. Feel free to change if you disagree :)


andre-meyer:
@mmvsha Please also create a PR for Master so we can update staging

PR #348, opened 2018-05-22
Task: Disable Analytics, abstracted into config entries

andre-meyer:
### Description
* Moved analytics into it's own folder and implement a generic loader pattern to load analytics/thirdparty integrations on runtime instead of on load, because for example google analytics is very aggressive with their tracking.

* /

### Which PRs are linked to my PR?
* /

* Google Analytics behaviour should remain the same

*Enabled google analytics in config.json*
* Added the `ga` scope, added the js dynamically

*Disabled google analytics in config.json*
* No `ga` in scope, no errors on load, no `analytics.js` in network tab.

### Background Information
* As soon as analytics.js is included, before any kind of load, it will already drop cookies.

### Configuration Entries
`/config.json`
```
{
  ...
  "thirdparty": {
    "googleAnalytics": {
      "enabled": true
    }
  }
}
```


src/utils/features.js
```diff
@@ -19,3 +19,10 @@ export const getProviderIntegrationConfig = providerName =>
   configInterface.providers.options &&
   configInterface.providers.options[providerName.toUpperCase()]) || {}
 
+export const isThirdPartyIntegrationEnabled = thirdPartyName =>
```
mmvsha: why didn't you use optional chaining operator? 
```
config.thirdparty?.[thirdPrtyName]?.enabled
```
https://github.com/tc39/proposal-optional-chaining

mmv08 (approved):

src/utils/features.js
andre-meyer: Because VSCode doesn't like it 😉 Will do though

mmvsha: Not necessary if you don't like it, It's just a personal preference 😄

PR #392, opened 2018-06-05
Bug PM-655: Fix "Connect a wallet" alignment, fix buy shares validation

mmvsha:
### Description
* Fix "Connect a wallet" alignment
* Fix buy shares form validation: now it compares the amount you want to invest to your balance of market's collateral token

* Fixes bug/PM-655

* Cannot think of any.

* Checked the validation to be correct on trading/tournament interface

**BEFORE/AFTER:**

screen shot 2018-06-05 at 14 18 13
screen shot 2018-06-05 at 14 18 19

### DIFF
--- a/src/components/Header/Header.mod.scss
+++ b/src/components/Header/Header.mod.scss
@@ -90,6 +90,8 @@ $navbarHeight: 40px;
 
     &.right {
       margin-left: auto;
+      display: flex;
+      align-items: center;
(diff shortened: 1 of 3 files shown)


mmvsha:
@andre-meyer Why? I don't think we want to redeploy mainnet because of this. Maybe we can do it later? When dev is merged to master

PR #407, opened 2018-06-13
Bug PM-672: Balance doesn't update after buy/sell/redeem

mmvsha:
### Description
* Update few dependencies
* Fetch balance after buy/sell/redeem

* Fixes bug/PM-672

* Can't think of any

Bought and sold tokens, redeemed my winnings. Balance was updated immediately.


src/containers/HeaderContainer/store/actions.js
```diff
@@ -13,7 +13,7 @@ import { fetchTournamentUserData } from 'routes/Scoreboard/store'
  * @param {function} dispatch
  * @param {function} getState
  */
-const requestCollateralTokenBalance = account => (dispatch, getState) => {
+export const requestCollateralTokenBalance = account => (dispatch, getState) => {
```
andre-meyer: I don't think this is the best place for this. Can we maybe move this into something like `/store/actions/account/tokens`? wdyt?

mmvsha: @andre-meyer I agree

PR #416, opened 2018-06-19
BUG PM-681: When user logs out from metamask, mainnet address and reward info are still shown

mmvsha:
### Description
* Set mainnet address to null when user logs out from metamask

* Fixes bug/PM-681

* Can't think of any

Logged out/in with metamask like 3 times and it worked correctly (hid/showed the account, rewards info)


src/integrations/store/actions/index.js
```diff
@@ -24,6 +25,15 @@ export const logoutProvider = () => async (dispatch, getState) => {
 }
 
 export const runProviderUpdate = (provider, data) => async (dispatch, getState) => {
+  if (provider.constructor.providerName === WALLET_PROVIDER.METAMASK && isFeatureEnabled('tournament')) {
```
andre-meyer: @mmvsha Hmm, so when the user uses uPort, the address will remain?

andre-meyer: I think this part can be removed, it should always delete the mainnet address when the wallet is locked/closed

mmvsha: we don't set mainnet address for uPort users @andre-meyer 
how do you detect if the wallet is closed?

andre-meyer: This was resolved :) Feel free to merge

PR #460, opened 2018-07-23
BUG PM-726: Make sure time zone displayed is universal

mmv08:
### Description
* Convert market resolution date from utc to local and compare it agains local one
* Switch to 24hr

* Fixes bug/PM-726

* None

*Market's resolution date*
* I made sure it's shown in 24hr format


mmvsha:
@andre-meyer all good, displays the correct value

PR #469, opened 2018-08-02
Bug PM-738: Scoreboard styles are not imported

mmvsha:
### Description
* Previously scoreboard's styles weren't imported and scoreboard looked weird

* Fixes bug/PM-738

* None

Checked that scoreboard shows fine

### Before and after:
screen shot 2018-08-02 at 16 14 02
screen shot 2018-08-02 at 16 13 23


src/containers/App/index.js
```diff
@@ -21,6 +21,10 @@ import style from './app.scss'
 import transitionStyles from './transitions.scss'
 import 'rc-tooltip/assets/bootstrap.css?raw'
 
+if (isFeatureEnabled('tournament')) {
```
mmvsha: @andre-meyer This is a good example why we should split configuration into runtme/build time

PR #476, opened 2018-08-15
BUG PM-720:  Add a GA opt-in banner

mmvsha:
### Description
* Add opt-in cookie banner
* Remove global styles for `label`
* Update dependencies versions

* Fixes task/PM-720

* Cant think of any

Tried different variations of cookies (ga selected only, intercom selected only, none is selected) and everything worked as expected.

### Configuration Entries
`/olympia/staging.json`
```
  "thirdparty": {
    "googleAnalytics": {
      "enabled": true,
      "id": "UA-83220550-2"
    },
    "intercom": {
      "enabled": true,
      "id": "km54f6ih"
    }
  },
  "cookieBanner": {
    "enabled": true,
    "options": [
      {
        "label": "Chat support",
        "thirdParty": "intercom"
      },
      {
        "label": "Analytics",
        "thirdParty": "googleAnalytics"
      }
    ]
  },
```


src/components/CookieBanner/style.scss
```diff
@@ -0,0 +1,121 @@
+@import '~style/vars.scss';
+
+$bgColor: #21435a;
+$textColor: #ffffff;
+$acceptButtonColor: #23d160;
```
andre-meyer: Can you move these variables into vars.scss and maybe try to use existing css vars?

src/components/CookieBanner/utils/getCookie.js
```diff
@@ -0,0 +1,6 @@
+const getCookie = (cname) => {
+  const match = document.cookie.match(`(^|;) ?${cname}=([^;]*)(;|$)`)
+  return match ? match[2] : ''
```
andre-meyer: Maybe we should use a small library to interact with cookies? Worried this could break in certain edge cases

src/components/CookieBanner/utils/setCookie.js
```diff
@@ -0,0 +1,13 @@
+const setCookie = (key, value, days) => {
+  const date = new Date()
+
+  let expirationDate
+  if (days) {
+    // Get unix milliseconds at current time plus number of days
+    expirationDate = new Date(date.setTime(+date + days * 86400000)) // 24 * 60 * 60 * 1000
+  }
+
+  document.cookie = `${key}=${value};expires=${expirationDate ? expirationDate.toUTCString() : ''}; path=/`
```
andre-meyer: Same here

src/containers/CookieBannerContainer/CookieBannerContainer.js
```diff
@@ -0,0 +1,50 @@
+import React, { Component } from 'react'
+import autobind from 'autobind-decorator'
+import CookieBanner from 'components/CookieBanner'
+import { THIRD_PARTY_INTEGRATIONS } from 'utils/analytics'
+import { getFeatureConfig } from 'utils/features'
+
+const { options } = getFeatureConfig('cookieBanner')
+const ThirdPartyIntegrations = options.map(integration => ({
+  ...integration,
+  initFunc: THIRD_PARTY_INTEGRATIONS[integration.thirdParty],
```
andre-meyer: It might be a good idea to ensure this doesn't crash if somebody passes an unknown thirdparty integration, when people run their own interface. `console.warn('unknown integration ${intergration.thirdParty}'` or something. Wdyt?

webpack.dev.config.js
```diff
@@ -177,6 +177,11 @@ module.exports = (env = {}) => {
             '@gnosis.pm/pm-js',
           ],
         },
+        plugins: [
+          new webpack.EnvironmentPlugin({
+            NODE_ENV: 'development',
```
andre-meyer: Is this also needed in webpack.prod?

src/components/CookieBanner/utils/getCookie.js
andre-meyer: https://www.npmjs.com/package/js-cookie

webpack.dev.config.js
mmvsha: @andre-meyer No, for some reason, it minifies the code by default. This is done to prevent minification

src/components/CookieBanner/style.scss
mmvsha: @andre-meyer why? It's not global variables, they're used only in this component

src/containers/CookieBannerContainer/CookieBannerContainer.js
mmvsha: @andre-meyer initFunc is market as required in CookieBanner's propTypes, so if its undefined there'll be an error message

PR #486, opened 2018-09-03
BUG PM-767: DecimalValue proptype, LegalCompliance fixes, Invalid winning outcome fix 

mmvsha:
### Description
* Fix `LegalCompliance` key warnings, also fix Text Explanation generation algorithm (previously " and " was prepended before the document list)
* Fix `Invalid Outcome` in `WinningOutcome` component
* Fix `DecimalValue` decimal.js proptype test
* Fix `CheckWalletConnection`, it would return true if tournament is not enabled but the user was registered previously (mainnetAddress stored in localStorage)
* Code style fixes 

* Fixes bug/PM-767

* Can't think of any, maybe some new interesting bugs

I made sure that:

* LegalCompliance doesn't show any warnings/errors
* Winning Outcome is shown correctly in scalar/categorical
* DecimalValue doesn't show PropType warnings anymore
* My tokens section is hidden if you're registered for tournament but not accepted legal documents

### DIFF
--- a/src/components/DecimalValue/index.js
+++ b/src/components/DecimalValue/index.js
@@ -44,8 +44,8 @@ export const decimalJsTest = (props, propName, componentName) => {
     namesToCheck.push(value.constructor.prototype.name)
   }
 
-  const testResults = namesToCheck.map(name => /^(Decimal|(Big)?Number)$/.test(name))
-  console.log(testResults)
+  const testResults = namesToCheck.map(name => /(Decimal|(Big)?Number)/.test(name))
+
   return testResults.indexOf(true) > -1
     ? undefined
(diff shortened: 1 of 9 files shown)


src/components/DecimalValue/index.js
```diff
@@ -23,20 +25,30 @@ export const decimalToText = (value, decimals = 4) => {
...
+
+  if (value.constructor) {
+    namesToCheck.push(value.constructor.name)
+    namesToCheck.push(value.constructor.prototype.name)
   }
-  return undefined
+
+  const testResults = namesToCheck.map(name => /(Decimal|(Big)?Number)/.test(name))
+
+  return testResults.indexOf(true) > -1
```
andre-meyer: use `.includes`?

